PR #12, opened 2023-04-21
Improve metrics to be available for more people

LeaveMyYard:
Some people do not see any CPU data because the metric calculation is not available. Reworked for a better one.


nick-dice:
Makes krr work for me, thanks.

robusta_krr/core/integrations/prometheus/loader.py
```diff
@@ -17,6 +15,8 @@
 from robusta_krr.utils.configurable import Configurable
 from robusta_krr.utils.service_discovery import ServiceDiscovery
 
+from .metrics import BaseMetricLoader
```
Avi-Robusta, line 18: fix relative import
`from robusta_krr.core.integrations.prometheus.metrics import BaseMetricLoader`
not sure how critical this is since I see you do it in a bunch of places, you know convention better than me

LeaveMyYard: We were replacing relative imports in robusta as we had `from .......src.core import something` 💀 
Relative imports from `.` or `..` can be fine, if those imports are about the linked module, like here: `BaseMetricLoader` is related to all other loaders

robusta_krr/core/integrations/prometheus/metrics/memory_metric.py
```diff
@@ -0,0 +1,60 @@
...
+                return series["metric"][label]
+        return None
+
+    @staticmethod
+    def filter_prom_jobs_results(
+        series_list_result: list[PrometheusSeries],
+    ) -> list[PrometheusSeries]:
+        """
+        Because there might be multiple metrics with the same name, we need to filter them out.
+
```
Avi-Robusta, line 26: I would comment why we do it here and not for CPU
Cpu usually only comes from node-exporter job

robusta_krr/core/integrations/prometheus/metrics/base_metric.py
```diff
@@ -0,0 +1,82 @@
...
+        self.prometheus = prometheus
+
+    @abc.abstractmethod
+    def get_query(self, namespace: str, pod: str, container: str) -> str:
+        ...
+
+    async def query_prometheus(
+        self, query: str, start_time: datetime.datetime, end_time: datetime.datetime, step: datetime.timedelta
+    ) -> list[dict]:
+        return await asyncio.to_thread(
```
Avi-Robusta, line 32: what happens when you get PrometheusNotFound?

Avi-Robusta (approved):
looks good, had a few questions
Looks familiar to what we did in runner

robusta_krr/core/integrations/prometheus/metrics/base_metric.py
LeaveMyYard: Here `PrometheusNotFound` would have been already raised. We have `self.prometheus` - an already working connection

PR #13, opened 2023-04-21
Add info on using port-forwarding for prometheus

LeaveMyYard:
Added a small section with an example of how to use port-forwarding to connect to an arbitrary prometheus

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,12 +247,12 @@ python krr.py simple --help
 
 If your prometheus is not auto-connecting, you can use `kubectl port-forward` for manually forwarding Prometheus.
 
-For example, if you have prometheus Pod called `kube-prometheus-st-prometheus-0`, then run this command to port-forward it:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -241,12 +241,33 @@ python krr.py simple --help
...
+## Example of using port-forward for Prometheus
+
+If your prometheus is not auto-connecting, you can use `kubectl port-forward` for manually forwarding Prometheus.
+
+For example, if you have prometheus Pod called `kube-prometheus-st-prometheus-0`, then run this command to port-forward it:
+```sh
+kubectl port-forward pod/kube-prometheus-st-prometheus-0 9090
+```
+
+Then, run another terminal and run robusta in it, adding explicit prometheus url:
```
aantn, line 255: Suggest a slight change for clarity:

*Then, open another terminal and run krr in it, giving an explicit prometheus url:*

aantn (approved):
Nice. Left two tiny comments.

PR #16, opened 2023-04-23
Try getting ingress before service LB

yonahd:


LeaveMyYard:
Looks good for me. @arikalon1 could you also take a look? Maybe we should add the same behaviour to Robusta?

LeaveMyYard (approved):

yonahd:
Hi @arikalon1 
When running internally, with tools such as robusta I think the service should be first(if not the only one) since the ingress is external.

But when running as a CLI the more common way to expose a service out of the cluster afaik is ingress which is why I added https://github.com/robusta-dev/krr/pull/16/files#diff-fd3b933f49f9ca8c088de045530c353275b8e51d6eadc6a00340d48ed1cd8e98R46

arikalon1:
Thanks for the clarification @yonahd 

With KRR we use the k8s proxy api to access internal prometheus services automatically discovered.
If it's ok with you, can you change the order, so that first we check if there's an internal service, and if it doesn't exist we will try to search for the ingress?

yonahd:
@arikalon1 the change you want is the service discovery to be before the ingress?

arikalon1:
@yonahd if that can work, yes

arikalon1:
Thanks for your contribution @yonahd

PR #33, opened 2023-05-10
Find older pods data using PromQL

LeaveMyYard:
Use PromQL to find all pods, related to the K8SObject.


LeaveMyYard:
Approved by @arikalon1 in https://github.com/robusta-dev/krr/pull/35

PR #35, opened 2023-05-11
Replace decimal to numpy and save time data

LeaveMyYard:
Loading additional data for older pods resulted in data, passed to strategy missing timestamp data, so no information about time relativity is preserved.

Additionally, replaces all Decimals to numpy.ndarray to speedup calculations and to get a cleaner calculations.

Merge only after https://github.com/robusta-dev/krr/pull/33


arikalon1 (approved):
LGTM

I think the code of the promql queries enhancements is also in this branch

PR #37, opened 2023-05-16
Shorter exception frames when kubewatch has an exception

Avi-Robusta:


LeaveMyYard (approved):
Fixed verbosity a bit, exception fix is nice

PR #38, opened 2023-05-16
Improve explainability: Add strategy description

LeaveMyYard:
Added the ability to provide docstrings to strategy class.

Those docstrings support [Rich Markup](https://rich.readthedocs.io/en/stable/markup.html#console-markup) and python string formatting.


Avi-Robusta (approved):
looks good, no changes needed

PR #40, opened 2023-05-17
Loading bar for calculating strategy

Avi-Robusta:


robusta_krr/core/runner.py
```diff
@@ -140,7 +139,8 @@ async def _collect_result(self) -> Result:
                 self.warning("Note that you are using the '*' namespace filter, which by default excludes kube-system.")
             return Result(scans=[])
 
-        resource_recommendations = await self._gather_objects_recommendations(objects)
+        with ProgressBar(self.config, total=len(objects), title="Calculating strategy") as progress_bar:
```
LeaveMyYard, line 142: Maybe to set here `progress_bar` as `self.__progressbar`, so there is no need to pass it through functions.

LeaveMyYard (changes requested):

robusta_krr/core/runner.py
LeaveMyYard: Maybe also to set a title as `"Calculating Recommendation"`

robusta_krr/core/runner.py
```diff
@@ -140,8 +139,10 @@ async def _collect_result(self) -> Result:
                 self.warning("Note that you are using the '*' namespace filter, which by default excludes kube-system.")
             return Result(scans=[])
 
+        self.__progressbar = ProgressBar(self.config, total=len(objects), title="Calculating Recommendation")
```
LeaveMyYard, line 142: you can use `self.something` in context managers

LeaveMyYard (approved):

PR #41, opened 2023-05-17
Add PromQL to json for Robusta UI

LeaveMyYard:
Combine promql requests for different pods into one, pass query to result data

### DIFF
--- a/robusta_krr/core/abstract/strategies.py
+++ b/robusta_krr/core/abstract/strategies.py
@@ -29,7 +29,7 @@ class StrategySettings(pd.BaseModel):
     history_duration: float = pd.Field(
         24 * 7 * 2, ge=1, description="The duration of the history data to use (in hours)."
     )
-    timeframe_duration: float = pd.Field(15, ge=1, description="The step for the history data (in minutes).")
+    timeframe_duration: float = pd.Field(2, ge=1, description="The step for the history data (in minutes).")
 
     @property
     def history_timedelta(self) -> datetime.timedelta:
--- a/robusta_krr/core/abstract/strategies.py
+++ b/robusta_krr/core/abstract/strategies.py
@@ -43,7 +43,16 @@ class StrategySettings(pd.BaseModel):
 _StrategySettings = TypeVar("_StrategySettings", bound=StrategySettings)
 
 ArrayNx2 = Annotated[NDArray[np.float64], Literal["N", 2]]
-ResourceHistoryData = dict[str, ArrayNx2]
+
+
+class ResourceHistoryData(pd.BaseModel):
+    query: str  # The query used to get the data
+    data: dict[str, ArrayNx2]  # Mapping: pod -> (time, value)
+
+    class Config:
+        arbitrary_types_allowed = True
+
+
 HistoryData = dict[ResourceType, ResourceHistoryData]
 RunResult = dict[ResourceType, ResourceRecommendation]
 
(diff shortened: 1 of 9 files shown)


robusta_krr/core/integrations/prometheus/metrics/cpu_metric.py
```diff
@@ -2,9 +2,17 @@
...
 from .base_filtered_metric import BaseFilteredMetricLoader
+from robusta_krr.core.models.objects import K8sObjectData
 
 
 @bind_metric(ResourceType.CPU)
 class CPUMetricLoader(BaseFilteredMetricLoader):
-    def get_query(self, namespace: str, pod: str, container: str) -> str:
-        return f'sum(irate(container_cpu_usage_seconds_total{{namespace="{namespace}", pod="{pod}", container="{container}"}}[5m])) by (container, pod, job)'
+    def get_query(self, object: K8sObjectData) -> str:
+        pods_selector = "|".join(pod.name for pod in object.pods)
```
Avi-Robusta, line 11: much better

Avi-Robusta: if we do this
 (
            'sum(irate(container_cpu_usage_seconds_total{}[5m])) by (container, pod, job)'
        )
     
     we will get all the metrics for every container and pod and can filter it on our end, it might be more efficient /  faster than calling query_prometheus for every deployment,DS,ss, job ..., WDYT?
     (same with memory)

Avi-Robusta: then again it might be allot of data that we get in the return result, it might be worth checking if the tradeoff is worth it

robusta_krr/core/abstract/strategies.py
```diff
@@ -43,7 +43,16 @@ def timeframe_timedelta(self) -> datetime.timedelta:
...
 ArrayNx2 = Annotated[NDArray[np.float64], Literal["N", 2]]
-ResourceHistoryData = dict[str, ArrayNx2]
+
+
+class ResourceHistoryData(pd.BaseModel):
+    query: str  # The query used to get the data
+    data: dict[str, ArrayNx2]  # Mapping: pod -> (time, value)
+
+    class Config:
+        arbitrary_types_allowed = True
```
Avi-Robusta, line 53: where is this used?

Avi-Robusta:
looks good, i would run a test comparing the table of the previous version to this to make sure nothing changed with the prometheus query

Avi-Robusta (approved):

robusta_krr/core/integrations/prometheus/metrics/cpu_metric.py
LeaveMyYard: Even tho it might be more efficient, in fact it currently is fast (mostly because we are doing that in parallel)
Also, we need the query to represent the resulted data for making graphs in Robusta UI.

robusta_krr/core/abstract/strategies.py
LeaveMyYard: This is because we have numpy type annotations inside pydantic, that is not supported. So we need to say to pydantic, that we can have some arbitraty types.

LeaveMyYard:
> some comments, looks good, i would run a test comparing the table of the previous version to this to make sure nothing changed with the prometheus query

Actually I got some rows to differ, I will investigate further

PR #55, opened 2023-05-29
Non-centralized vm and thanos support

Avi-Robusta:


robusta_krr/core/integrations/prometheus/loader.py
```diff
@@ -92,65 +38,31 @@ def __init__(
...
+            cluster: Optional[str] = None,) -> Optional[MetricsService]:
+        
+        for service_name, metric_service_class in METRICS_SERVICES.items():
+            try:
+                loader = metric_service_class(config, api_client=api_client, cluster=cluster)
+                loader.check_connection()
+                self.echo(f"{service_name} found")
+                return loader
+            except Exception as e:
+                self.warning(f"{service_name} not found")
```
LeaveMyYard, line 65: I'm not sure this should be a warning, rather the debug. Like: why should the used be warned that we can not find victoriametrics, if they are using thanos?

robusta_krr/core/integrations/prometheus/metrics_service/thanos_metrics_service.py
```diff
@@ -0,0 +1,64 @@
...
+            selectors=[
+                "app.kubernetes.io/component=query,app.kubernetes.io/name=thanos",
+                "app.kubernetes.io/name=thanos-query",
+                "app=thanos-query",
+                "app=thanos-querier",
+            ]
+        )
+
+
+class ThanosMetricsNotFound(Exception):
```
LeaveMyYard, line 30: Should it be a subclass of `PrometheusNotFound`? Or maybe they should subclass some common parent aka `MetricsNotFound`? 

Because maybe we will want to catch all of them in one try block

robusta_krr/core/integrations/prometheus/metrics_service/base_metric_service.py
```diff
@@ -0,0 +1,35 @@
...
+        super().__init__(config=config)
+        self.api_client = api_client
+        self.cluster = cluster or 'default'
+    
+    @abc.abstractmethod
+    def check_connection(self):
+        ...
+
+    @abc.abstractmethod
+    def name(self) -> str:
```
LeaveMyYard, line 23: Add a default implementation, so you do not have to override it everywhere (wrote without checking, verify it is correct):

```python
def name(self) -> str:
    classname = self.__class__.__name__
    return classname[:-14] if classname.lower().endswith("metricsservice") else classname
```

LeaveMyYard (changes requested):
Also please run `black . && isort . && flake8 .`

robusta_krr/core/integrations/prometheus/metrics_service/thanos_metrics_service.py
Avi-Robusta: sounds good

robusta_krr/core/integrations/prometheus/metrics_service/base_metric_service.py
Avi-Robusta: you are the python king

Issue #66, opened 2023-06-09
PrometheusApiClientException: HTTP Status Code 414 Request-URI Too Large


infa-ddeore:
**Describe the bug**
"krr simple -n kube-system -p https://prometheus.example.com" gives "Request-URI Too Large" exception

**To Reproduce**
not sure if this is reproducible in every environment, i am trying the command on a cluster having 165 nodes, not sure if that is a big number for `krr`

**Expected behavior**
report should appear

**Screenshots**
pasting the error with masked url and cluster name
```
$ krr simple -n kube-system -p https://prometheus.example.com


 _____       _               _          _  _______  _____
|  __ \     | |             | |        | |/ /  __ \|  __ \
| |__) |___ | |__  _   _ ___| |_ __ _  | ' /| |__) | |__) |
|  _  // _ \| '_ \| | | / __| __/ _` | |  < |  _  /|  _  /
| | \ \ (_) | |_) | |_| \__ \ || (_| | | . \| | \ \| | \ \
|_|  \_\___/|_.__/ \__,_|___/\__\__,_| |_|\_\_|  \_\_|  \_\



Running Robusta's KRR (Kubernetes Resource Recommender) v1.2.1
Using strategy: Simple
Using formatter: table

[INFO] Using clusters: ['k8-cluster']
[INFO] Listing scannable objects in k8-cluster
[INFO] Found 30 objects across 1 namespaces in k8-cluster
on 0: [INFO] Connecting to Prometheus for k8-cluster cluster
on 0: [INFO] Using Prometheus at https://prometheus.example.com for cluster k8-cluster
on 0: [INFO] Prometheus found
on 0: [INFO] Prometheus connected successfully for k8-cluster cluster
Calculating Recommendation |⚠︎                                       | (!) 0/30 [0%] in 1:55.2 (0.00/s)
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /private/tmp/robusta_krr/core/runner.py:174 in run                                               │
│                                                                                                  │
│ [Errno 2] No such file or directory: '/private/tmp/robusta_krr/core/runner.py'                   │
│                                                                                                  │
│ /private/tmp/robusta_krr/core/runner.py:153 in _collect_result                                   │
│                                                                                                  │
│ [Errno 2] No such file or directory: '/private/tmp/robusta_krr/core/runner.py'                   │
│                                                                                                  │
[...]

Issue #62, opened 2023-06-05
Restrict workload selection


dohnto-s1:
**Is your feature request related to a problem? Please describe.**
One might have a huge cluster and want to focus on optimizing just a particular workload.
Prometheus queries while analysing all worklaods in a cluster/namespace might take too long otherwise.

**Describe the solution you'd like**
Introduce a new command line argument as kubectl has:

```
    -l, --selector='':
	Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2). Matching
	objects must satisfy all of the specified label constraints.
```

to limit the workload selection. E. g:
```
krr simple --context my-cluster --namespace kube-system --selector app-instance=metrics-server
# or
krr simple --context my-cluster --namespace kube-system --selector owner=devops
```

**Describe alternatives you've considered**

```
krr simple --context my-cluster --namespace kube-system --workload Deployment/metrics-server
```

**Additional context**

Thanks for this wonderful tool. Happy to learn if there is alternative solution for me.

dohnto-s1:
Now I hit an issue when Prometheus is unable to give me the results and the request fails with 431. I don't understand the details `krr`, but would implementation of feature above help as well? 

> PrometheusApiClientException: HTTP Status Code 431 (b'431 Request Header Fields Too Large')

arikalon1:
Thanks for reporting it @dohnto-s1 

I think this issue is similar to #60 

Does limiting `krr` to a specific namespace using `krr simple -n default` can be a workaround until we fix it?

dohnto-s1:
This unfortunately happens when running with `-n`.

LeaveMyYard:
And I really like your proposal for selector flag, we should 100% implement it

dohnto-s1:
@LeaveMyYard Looks awesome, thanks!

PR #67, opened 2023-06-10
Reduce size of pods_selector by using replica prefixes

### LINKED ISSUE #60: Too many historic pods causes querystring too long for Prometheus range_query API
**Describe the bug**

If we have many historic pods, the query expr will be very long. Then at some point, the request may be rejected by the gateway, e.g. Nginx. HTTP code 422 may be returned.

**To Reproduce**

**Expected behavior**

Use POST API instead of GET.

https://github.com/4n4nd/prometheus-api-client-python/blob/39c5710521134fc450e9b4103cbb5995c05c5273/prometheus_api_client/prometheus_connect.py#L403-L409

But since we are using `prometheus-api-client-python`, it is not possible to do this.

**Screenshots**

image

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

lujiajing1126:
Closes #60, #62, #66

This PR intends to reduce the size of pod_selector, by picking the `<deployment>-<pod-template-hash>` as the prefix. 

This could significantly reduce the size of the querystring in the case that there are hundreds of pods for a single deployment.

`300` may be a magic number here: `pod-template-hash` normally contains 8 bytes, with 5 chars pod hash. The default max http header that nginx can accept is ~4K, so with >300 pods it must exceed the threshold.


LeaveMyYard:
Hello and sorry for long time waiting on response :) 

We've been working a lot with performance issues and also trying your solution.
One of the solutions we tried is to make pod gathering and metric gathering in one query:
https://github.com/robusta-dev/krr/blob/rework-promql-queries/robusta_krr/core/integrations/prometheus/metrics/cpu_metric.py

It resulted in using even more memory load on prometheus to execute that query, as well as longer loading time.
Currently we thing that the correct solution will be to use prometheus with POST except of GET to avoid the limitation of request URI being too large (https://github.com/robusta-dev/krr/issues/66)
You have mentioned that in https://github.com/robusta-dev/krr/issues/60, but still we can use POST by just subclassing the lib we are using for prometheus

Additionally, I've been thinging on fully rewriting regex to fully utilize the format for pod names, but there is a concern that there might be special cases where it fails.

In any case thank you so much for

LeaveMyYard:
I am closing this PR together with your issue as https://github.com/robusta-dev/krr/pull/97 should have fixed the issue

Still, reopen it if you feel like it improves anything else, but I will keep this possibility in mind in any case

Thank you very much for participating in a project ❤️

lujiajing1126:
> Hello and sorry for long time waiting on response :)
> 
> We've been working a lot with performance issues and also trying your solution. One of the solutions we tried is to make pod gathering and metric gathering in one query: https://github.com/robusta-dev/krr/blob/rework-promql-queries/robusta_krr/core/integrations/prometheus/metrics/cpu_metric.py
> 
> It resulted in using even more memory load on prometheus to execute that query, as well as longer loading time. 

I've tested this solution on VictoriaMetrics (a cluster w/ 4 vmstorage), but it failed to return data and throws timeout. Probably it is too complicated.

I don't know if anyone else can make another test.

> Currently we thing that the correct solution will be to use prometheus with POST except of GET to avoid the limitation of request URI being too large (#66) You have mentioned that in #60, but still we can use POST by just subclassing the lib we are using for prometheus

Sure. I forgot to mention this. We've modified site-package to make use of POST. But besides that, vmselect has an argument,

```
  -search.maxQueryLen size
     The maximum search query length in bytes
     Supports the following optional suffixes for size values: KB, MB, GB, TB, KiB, MiB, GiB, TiB (default 16384)
```

The maximum query length is 16KiB. So we still have a limit even for POST.

PR #68, opened 2023-06-12
fix bug only collect one pod series per container

Avi-Robusta:
Bug explained
we query for metric A on container B for pods X,Y,Z
in function `filter_prom_jobs_results` the bug caused us to only process the results of one pod instead of all of the relevant pods


robusta_krr/core/integrations/prometheus/metrics/base_filtered_metric.py
```diff
@@ -17,7 +17,7 @@ class BaseFilteredMetricLoader(BaseMetricLoader):
 
     @staticmethod
     def get_target_name(series: PrometheusSeries) -> Optional[str]:
-        for label in ["container", "pod", "node"]:
+        for label in [ "pod","container", "node"]:
```
LeaveMyYard, line 20: Run black, please

LeaveMyYard (approved):

PR #69, opened 2023-06-13
Requests diff in table view

LeaveMyYard:
Add difference and total difference in table formatter

### DIFF
--- a/robusta_krr/formatters/table.py
+++ b/robusta_krr/formatters/table.py
@@ -21,17 +21,25 @@ def _format(value: RecommendationValue) -> str:
         return resource_units.format(value)
 
 
+def __calc_diff(allocated, recommended, selector, multiplier=1) -> str:
+    if recommended is None or isinstance(recommended, str) or selector != "requests":
+        return ""
+    else:
+        reccomended_val = recommended.value if isinstance(recommended.value, (int, float)) else 0
+        allocated_val = allocated if isinstance(allocated, (int, float)) else 0
+        diff_val = reccomended_val - allocated_val
+        diff_sign = "[green]+[/green]" if diff_val >= 0 else "[red]-[/red]"
+        return f"{diff_sign}{_format(abs(diff_val) * multiplier)}"
+
+
 def _format_request_str(item: ResourceScan, resource: ResourceType, selector: str) -> str:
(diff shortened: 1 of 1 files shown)


robusta_krr/formatters/table.py
```diff
@@ -20,12 +21,40 @@ def _format(value: RecommendationValue) -> str:
         return resource_units.format(value)
 
 
+def __calc_diff(allocated, recommended, selector, multiplier=1) -> str:
+    if recommended is None or isinstance(recommended, str) or selector != "requests":
```
Avi-Robusta, line 25: you only want to calculate the difference for requests and not limits?

robusta_krr/formatters/table.py
```diff
@@ -47,15 +76,19 @@ def table(result: Result) -> Table:
...
+    if cluster_count > 1:
+        table.add_column("Cluster", style="cyan")
     table.add_column("Namespace", style="cyan")
     table.add_column("Name", style="cyan")
     table.add_column("Pods", style="cyan")
     table.add_column("Old Pods", style="cyan")
     table.add_column("Type", style="cyan")
     table.add_column("Container", style="cyan")
     for resource in ResourceType:
+        table.add_column(f"{resource.name} Diff")
```
Avi-Robusta, line 91: if you are only calculating the difference for requests and displaying it there isnt this column redundant?

robusta_krr/formatters/table.py
LeaveMyYard: Yes, [here](https://docs.google.com/spreadsheets/d/1NUhBiKzsPioI8xyAOzDej3C7bomjPQnVJTb0izCuJcQ/edit#gid=1035852509) is the link for the spec

robusta_krr/formatters/table.py
Avi-Robusta: nvrmnd saw that the multiple pods its different

Avi-Robusta (approved):
looks good, 
Personally not in love with the diff column but if its what they want than dont worry about it

robusta_krr/formatters/table.py
Avi-Robusta: cool, just wanted to make sure, was confused that i didnt see it for limits

PR #70, opened 2023-06-13
Centralized multi-cluster Prometheus/Thanos/VM support

Avi-Robusta:
we require the user to add the label when running krr on multi-cluster centeralized prometheus 
example of the new command:
`python3.9 krr.py simple -p http://PROM_URL:9090 -l prod-eu-cluster`
When a label is not specified you get this error:
`[ERROR] KRR requires a prometheus_cluster_label to run your query, please select one of the following clusters ['CLUSTER_1_NAME', 'CLUSTER_2_NAME']`

if the cluster label doesn't exist you get the following error:
`[ERROR] prometheus_cluster_label bad-cluster-name does not exist, please use one of the following ['CLUSTER_1_NAME', 'CLUSTER_2_NAME']`


aantn:
Avi, can we explicitly name the command line parameter in the error
message. E.g:

Rerun krr with the flag `-l <cluster>` where <cluster> is one of...

LeaveMyYard (approved):

PR #71, opened 2023-06-14
multi-cluster support, custom label name

Avi-Robusta:
the standard label name is usually clusters but theoretically it can be anything so we support that


robusta_krr/core/integrations/prometheus/metrics/base_metric.py
```diff
@@ -39,7 +39,7 @@ def get_prometheus_cluster_label(self) -> str:
         """
         if self.config.prometheus_cluster_label is None:
             return ""
-        return f', cluster="{self.config.prometheus_cluster_label}"'
+        return f', {self.config.prometheus_label}="{self.config.prometheus_cluster_label}"'
```
LeaveMyYard, line 42: It seems that `prometheus_label` is optional. but you use it here (it will be `", None=..."`)

LeaveMyYard (approved):

robusta_krr/core/integrations/prometheus/metrics/base_metric.py
Avi-Robusta: good call, it has a default value so it will never be none

PR #75, opened 2023-06-20
Slack Integration out of the box

arnoldyahad:
Hey guys,
We are using this great tool and wanted to have the report sent to slack.
We noticed the slack integration is not part of the KRR but part of robusta, as we couldn't find any way to run this tool and have the output being sent to a channel of our choice. 

We decided to develop a solution so it will be easily sent to slack using the krr itself:
```
krr simple --slackoutput krr-report-test
```

In this example, `krr-report-test` is the target channel.

This is how the result looks like:
image

when using the help option it will present the new options:
image

a requirement for this is the `SLACK_BOT_TOKEN` (name taken from slack documentation) that is needed to authenticate with slack, if you already have robusta in your cluster you can use the same token as it needs the same permissions.

This code has been written very quickly and its far from being production ready - but if putting it in a PR in case someone from the community would like to use it or to "perfect" it so it can be merged.


robusta_krr/main.py
```diff
@@ -108,6 +108,8 @@ def {func_name}(
                 verbose: bool = typer.Option(False, "--verbose", "-v", help="Enable verbose mode", rich_help_panel="Logging Settings"),
                 quiet: bool = typer.Option(False, "--quiet", "-q", help="Enable quiet mode", rich_help_panel="Logging Settings"),
                 log_to_stderr: bool = typer.Option(False, "--logtostderr", help="Pass logs to stderr", rich_help_panel="Logging Settings"),
+                file_output: Optional[str] = typer.Option(None, "--fileoutput", help="Print to file", rich_help_panel="Output Settings"),
+                slack_output: Optional[str] = typer.Option(None, "--slackoutput", help="Send to slack", rich_help_panel="Output Settings"),
```
LeaveMyYard, line 112: Here also. Tell that slack output needs an Env variable to be set with a token.

LeaveMyYard (changes requested):
I think you forgot to add slack_sdk to libraries required.

Also maybe this output feature should be documented in readme

Also, run `black` on the code to make it a bit better

Overall good improvement, thanks for helping the project :)

PR #78, opened 2023-06-21
Add strategy data in raw result

LeaveMyYard:
Add strategy information, specific for UI scan description (see https://github.com/robusta-dev/robusta/pull/941)


Avi-Robusta (approved):
looks good

PR #80, opened 2023-06-22
Hpa usage in calculation

LeaveMyYard:
Load HPA data if availble. Skip calculation for the resource if the HPA is available

TODO: Save message info data to database for Robusta UI

### DIFF
--- a/robusta_krr/core/integrations/kubernetes.py
+++ b/robusta_krr/core/integrations/kubernetes.py
@@ -115,15 +115,19 @@ class ClusterLoader(Configurable):
         return [PodData(name=pod.metadata.name, deleted=False) for pod in ret.items]
 
     async def __build_obj(self, item: AnyKubernetesAPIObject, container: V1Container) -> K8sObjectData:
+        name = item.metadata.name
+        namespace = item.metadata.namespace
+        kind = item.__class__.__name__[2:]
+
         return K8sObjectData(
             cluster=self.cluster,
-            namespace=item.metadata.namespace,
-            name=item.metadata.name,
-            kind=item.__class__.__name__[2:],
+            namespace=namespace,
+            name=name,
+            kind=kind,
             container=container.name,
             allocations=ResourceAllocations.from_container(container),
             pods=await self.__list_pods(item),
-            hpa=self.__hpa_list.get((item.kind, item.metadata.name)),
+            hpa=self.__hpa_list.get((kind, name)),
         )
 
     async def _list_deployments(self) -> list[K8sObjectData]:
--- a/robusta_krr/core/integrations/kubernetes.py
+++ b/robusta_krr/core/integrations/kubernetes.py
@@ -1,11 +1,9 @@
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import itertools
-import os
 from typing import Optional, Union
 
 from kubernetes import client, config  # type: ignore
-from kubernetes.client.rest import ApiException
 from kubernetes.client.models import (
     V1Container,
     V1DaemonSet,
@@ -20,11 +18,7 @@ from kubernetes.client.models import (
     V1StatefulSet,
     V1StatefulSetList,
     V2HorizontalPodAutoscaler,
-    V2HorizontalPodAutoscalerSpec,
-    V2HorizontalPodAutoscalerBehavior,
-    V2MetricSpec,
-    V2ResourceMetricSource,
-    V2MetricTarget,
+    V2HorizontalPodAutoscalerList,
 )
 
 from robusta_krr.core.models.objects import K8sObjectData, PodData, HPAData
@@ -63,6 +57,7 @@ class ClusterLoader(Configurable):
         self.debug(f"Namespaces: {self.config.namespaces}")
 
         try:
+            self.__hpa_list = await self.__list_hpa()
             objects_tuple = await asyncio.gather(
                 self._list_deployments(),
                 self._list_all_statefulsets(),
@@ -128,7 +123,7 @@ class ClusterLoader(Configurable):
             container=container.name,
             allocations=ResourceAllocations.from_container(container),
             pods=await self.__list_pods(item),
-            hpa=await self._get_hpa(item),
+            hpa=self.__hpa_list.get((item.kind, item.metadata.name)),
         )
 
     async def _list_deployments(self) -> list[K8sObjectData]:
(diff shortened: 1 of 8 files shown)


robusta_krr/core/integrations/kubernetes.py
```diff
@@ -111,6 +122,7 @@ async def __build_obj(
             container=container.name,
             allocations=ResourceAllocations.from_container(container),
             pods=await self.__list_pods(item),
+            hpa=await self._get_hpa(item),
```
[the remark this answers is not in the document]
LeaveMyYard: I agree, just for some reason did not found this call myself

robusta_krr/strategies/simple.py
```diff
@@ -48,11 +48,24 @@ class SimpleStrategy(BaseStrategy[SimpleStrategySettings]):
     __display_name__ = "simple"
     __rich_console__ = True
 
-    def run(self, history_data: HistoryData, object_data: K8sObjectData) -> RunResult:
+    def __calculate_cpu_proposal(self, history_data: HistoryData, object_data: K8sObjectData) -> ResourceRecommendation:
+        if object_data.hpa is not None and object_data.hpa.target_cpu_utilization_percentage is not None:
+            return ResourceRecommendation.undefined()
```
Avi-Robusta, line 53: for the table for OOMKill it was requested to add a star by the reccomendation and at the bottom add a line like "*this container had oomkills"
I would consult tomer if something simmilar should be done for HPA
like saying "**this is run by HPA so the metrics can not be calculated"

robusta_krr/strategies/simple.py
```diff
@@ -48,11 +48,24 @@ class SimpleStrategy(BaseStrategy[SimpleStrategySettings]):
...
+
         cpu_usage = self.settings.calculate_cpu_proposal(history_data[ResourceType.CPU].data)
+        return ResourceRecommendation(request=cpu_usage, limit=None)
+
+    def __calculate_memory_proposal(
+        self, history_data: HistoryData, object_data: K8sObjectData
+    ) -> ResourceRecommendation:
+        if object_data.hpa is not None and object_data.hpa.target_memory_utilization_percentage is not None:
```
Avi-Robusta, line 61: I think this might be a bug.
For an HPA that is scaled down all the way we will still run memory recommendations
Screen Shot 2023-06-22 at 15 42 57

robusta_krr/core/models/objects.py
```diff
@@ -13,12 +13,22 @@ def __hash__(self) -> int:
         return hash(self.name)
 
 
+class HPAData(pd.BaseModel):
+    min_replicas: int
```
Avi-Robusta, line 17: I think this field might be optional
https://www.manifests.io/kubernetes/1.26/io.k8s.api.autoscaling.v1.HorizontalPodAutoscalerSpec?linked=HorizontalPodAutoscaler.spec

robusta_krr/core/models/objects.py
```diff
@@ -13,12 +13,22 @@ def __hash__(self) -> int:
...
+    max_replicas: int
+    current_replicas: int
```
Avi-Robusta, line 19: I could be wrong but I remember having an issue with this field being empty while the autoscaller is starting. It is manditory according to the docs Iv seen.
I would have this as optional too just to be safe so we dont have an exception in the rare case an autoscaller is starting up.

robusta_krr/core/runner.py
```diff
@@ -165,6 +175,10 @@ async def _collect_result(self) -> Result:
                 for obj, (recommended, metrics) in zip(objects, resource_recommendations)
             ],
             description=self._strategy.description,
+            strategy=StrategyData(
```
Avi-Robusta, line 178: this looks like its supposed to be from a different pr

robusta_krr/strategies/simple.py
LeaveMyYard: I don't understand. In your screenshot averageUtilization is `0`, not `null`. So we will not run the calculation

robusta_krr/core/runner.py
LeaveMyYard: Yeah, I did not checkout accidentally - but it will be removed after another merge

robusta_krr/strategies/simple.py
LeaveMyYard: Created a PR for the playbook also
https://github.com/robusta-dev/robusta/pull/953

robusta_krr/strategies/simple.py
Avi-Robusta: cpu averageUtilization is 0 not memory,
see how your function __get_metric(hpa, "memory") would run on this

robusta_krr/formatters/table.py
```diff
@@ -45,7 +46,13 @@ def _format_request_str(item: ResourceScan, resource: ResourceType, selector: st
...
 
     return (
-        diff + f"[{severity.color}]" + _format(allocated) + " -> " + _format(recommended.value) + f"[/{severity.color}]"
+        diff
+        + f"[{severity.color}]"
+        + _format(allocated)
+        + " -> "
+        + _format(recommended.value)
+        + f"[/{severity.color}]"
+        + (f" [grey27]({info})[/grey27]" if info else "")
```
Avi-Robusta, line 55: Do you have a screenshot of the changes?

Avi-Robusta (approved):
see what I wrote about the bug,
other than that look good, It doesn't look like it will impact performance on prometheus so I would merge when you fix the bug

robusta_krr/strategies/simple.py
LeaveMyYard: Recommendation for memory is expected when there is only hpa for CPU and vice-verca

erplsf:
why CPU calculation was skipped for HPAs? wondering before we start to implement a custom strategy to handle it if there's any deep architectural reasons why we can't handle HPA cpu the same way as non-HPA cpu.

aantn:
@erplsf there are some complexities so we decided to just skip for now. Here's an example scenario:

1. A pod has a request of 1 CPU before running Robusta.
2. HPA is configured to scale at 80% CPU.
3. Peak CPU will be around 0.8 CPU. (99th percentile can be underneath that depending on the usage pattern.)
4. KRR will recommend 0.8 CPU 
5. Assume the user applied the recommendations, and the request now is 0.8 CPU. The HPA wasn't changed and is still at 80% CPU.
6. Now, the peek will be 80% * 0.8, which is 0.64 CPU
7. Krr will recommend 0.64 CPU
8. ...

That said, it's possible to do this right! Please contribute back a custom strategy if you do this! We want to support the HPA with an HPA-aware strategy. Removing HPA support was a short-term decision until someone does it properly.


1. You have an HPA that targets 70% CPU utilization.
2. KRR outputs a recommendation based on the 95% percentile

erplsf:
thank you for the detailed description! I'll see if I can come up with something that makes sense (at least for our usecase), and if it's worth contributing upstream.

aantn:
Thanks, sounds good.

Keep in mind that however specific your use-case, it's almost certainly not unique! So feel free to discuss any solution here. I suspect it will end up being relevant to others too.

PR #82, opened 2023-06-22
[MAIN-553] Create explicit executors to limit threadcount

LeaveMyYard:
Replace `asyncio.to_thread` to `loop.run_in_executor` to explicitly provide executor.
Default python `max_thread_count` turned out to eat away lots of resources or even kill prometheus on big k8s nodes.

Note:
- Executor for runner running calculations when the data is collected is set to 6
- Executor for Kubernetes API fetcher is set to 6, but there is one for each cluster
- Executor for Prometheus API fetcher is set to 6, but there is one for each cluster


robusta_krr/core/runner.py
```diff
@@ -116,7 +121,8 @@ async def _calculate_object_recommendations(self, object: K8sObjectData) -> tupl
 
         # NOTE: We run this in a threadpool as the strategy calculation might be CPU intensive
         # But keep in mind that numpy calcluations will not block the GIL
-        result = await asyncio.to_thread(self._strategy.run, data, object)
+        loop = asyncio.get_running_loop()
+        result = await loop.run_in_executor(self._executor, self._strategy.run, data, object)
```
Avi-Robusta, line 125: do we want to implement some sort of rate limiter here also, there will still could be almost constant queries to prometheus

robusta_krr/core/integrations/kubernetes.py
```diff
@@ -26,6 +28,8 @@ def __init__(self, cluster: Optional[str], *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         self.cluster = cluster
+        # This executor will be running requests to Kubernetes API
+        self.executor = ThreadPoolExecutor(6)
```
Avi-Robusta, line 32: This could also make the scan take longer, krr_scan times out after 5 minutes, we might want to run some tests before we merge to get an idea for the time increase that we will have now or we could break the action

robusta_krr/core/integrations/kubernetes.py
```diff
@@ -190,7 +211,7 @@ async def list_clusters(self) -> Optional[list[str]]:
             return None
 
         try:
-            contexts, current_context = await asyncio.to_thread(config.list_kube_config_contexts)
+            contexts, current_context = config.list_kube_config_contexts()
```
Avi-Robusta, line 214: why did you remove the thread instead of doing something like this?
```
contexts, current_context = await loop.run_in_executor(
            self.executor, lambda: config.list_kube_config_contexts())
```

Avi-Robusta (changes requested):
Looks good, I had one thing I wanted clarity for.
Also I thought we discussed also adding rate limiting in the meeting too.
This could cause break on runner so I would like you to check the time it takes to run on average since by default runner waits only five minutes

robusta_krr/core/integrations/kubernetes.py
LeaveMyYard: Because this config listing is both fast and blocking, I descided that it does not need running in ThreadPool

robusta_krr/core/integrations/kubernetes.py
LeaveMyYard: So, by default `await asyncio.to_thread()` runs the syncronous task in a default threadpool, being limited to `min(32, os.cpu_count() + 4)`.

So, a node with lots of cores will have max of 32 threads opened.

I changed that behaviour to limit the size of a threadpool, so limiting the amount of requests is also provided that way.

robusta_krr/core/runner.py
LeaveMyYard: Same, we limited the query rate by limiting the executor

LeaveMyYard:
> Looks good, I had one thing I wanted clarity for. Also I thought we discussed also adding rate limiting in the meeting too. This could cause break on runner so I would like you to check the time it takes to run on average since by default runner waits only five minutes

I have tested the different multithreading settings and the result is:
Lower multithreading count results in less loading on prometheus memory usage, as well as slowing down the scan process (1 thread scanning 44 objects was working 1m11s).

For bigger count of threads the speed increased. 6 threads consumed more memory of prometheus then 32, but I think we need to have a bigger test with a bigger cluster (32+ cores for a node, 300+ pods).

I am providing a CLI argument to change the amount of max workers. If bigger tests show that 6 is still to much, we will be able to decrease the default value, but for now it should do the work.

LeaveMyYard:
Keep in mind that performance is soon to be optimized by decreasing the amount of queries to prometheus.

robusta_krr/core/integrations/kubernetes.py
LeaveMyYard: Async thread is waiting for `list_kube_config_contexts` to complete before continuing anything else (we await the `list_clusters`).
So it does not really matter if we run it in a threadpool

Avi-Robusta (approved):

PR #85, opened 2023-06-25
Rollout integration

arnoldyahad:
# Rollout Integration
If you are using Rollouts by ArgoCD instead of deployments - `krr` won't show it as it doesnt support rollouts.
since we are using `krr` and we are relying heavily on Rollouts - we created this PR in-order for `krr` to bring recommendations to rollouts as well.

To not make huge changes in the way `krr` works, I have added a change to the kubernetes client - added 2 functions:
 `list_rollout_for_all_namespaces`
 `list_rollout_for_all_namespaces_with_http_info`
Which are copies of the original functions of deployment, just for rollouts.

This is the most minimal way i've found to do this change and add support for rollouts (using any other custom object api by kubernetes resulted in it not being good enough for `V1DeploymentList` or any `V1CustomList`)

### DIFF
--- a/robusta_krr/core/integrations/kubernetes.py
+++ b/robusta_krr/core/integrations/kubernetes.py
@@ -20,6 +20,8 @@ from robusta_krr.core.models.objects import K8sObjectData, PodData
 from robusta_krr.core.models.result import ResourceAllocations
 from robusta_krr.utils.configurable import Configurable
 
+from .rollout import RolloutAppsV1Api
+
 
 class ClusterLoader(Configurable):
     def __init__(self, cluster: Optional[str], *args, **kwargs):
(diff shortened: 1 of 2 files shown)


LeaveMyYard (changes requested):
I like your PR, the only thing I would like to change it the location of the rollout file. 
I feel like I would have done a submodule in kubernetes folder, as rollout is not itself an integration, but rather an addition to kubernetes integration.

Overall code looks good, but I want to test it additionally before merging

LeaveMyYard:
Just checked and `_list_rollouts` fails with 404 not found in my cluster. Despite me having an ArgoCD installed, it seems that I do not have rollouts API. So that should be addressed also.

LeaveMyYard (approved):

PR #88, opened 2023-06-27
Rework promql queries

LeaveMyYard:


Avi-Robusta:
single thread memory in prometheus was 1.3gb 
6 threads crashed my node
Screen Shot 2023-07-02 at 17 47 22
its a small cluster (4 nodes 40 objects) so the preformance shouldnt be good, I would hold off merging until we fix the preformance issue

robusta_krr/core/integrations/prometheus/metrics/memory_metric.py
```diff
@@ -8,13 +8,29 @@
...
+                        container_memory_working_set_bytes{{namespace="{object.namespace}", container="{object.container}", image!=""}}
+                    )
+                )
+            """
+        else:
+            return f"""
+                max by (namespace, pod) (
+                    kube_pod_owner{{owner_kind="{object.kind}", owner_name="{object.name}"}}
+                ) * on(namespace, pod) group_right() avg by (namespace, pod, container) (
+                    sum by (namespace, pod, container, job, service) (
```
Avi-Robusta, line 32: we need to add id here too - or else it will think a container restarting will be a memory spike

robusta_krr/core/integrations/prometheus/metrics/memory_metric.py
```diff
@@ -8,13 +8,29 @@
...
+                max by (namespace, pod) (
+                    label_replace(
+                        kube_pod_owner{{owner_kind="ReplicaSet"}}, "replicaset", "$1", "owner_name", "(.*)"
+                    ) * on (namespace, replicaset) group_left(owner_name) topk by (namespace, replicaset) (
+                        1, max by (namespace, replicaset, owner_name) (
+                            kube_replicaset_owner{{owner_kind="Deployment", owner_name="{object.name}"}}
+                        )
+                    )
+                ) * on(namespace, pod) group_right() avg by (namespace, pod, container) (
+                    sum by (namespace, pod, container, job, service) (
```
Avi-Robusta, line 22: we need to add id here too - or else it will think a container restarting will be a memory spike

robusta_krr/core/integrations/prometheus/metrics/memory_metric.py
```diff
@@ -8,13 +8,29 @@
...
+                        container_memory_working_set_bytes{{namespace="{object.namespace}", container="{object.container}", image!=""}}
+                    )
+                )
+            """
+        else:
+            return f"""
+                max by (namespace, pod) (
+                    kube_pod_owner{{owner_kind="{object.kind}", owner_name="{object.name}"}}
+                ) * on(namespace, pod) group_right() avg by (namespace, pod, container) (
```
Avi-Robusta, line 31: why avg btw?

robusta_krr/core/models/objects.py
```diff
@@ -18,10 +18,10 @@ class K8sObjectData(pd.BaseModel):
     cluster: Optional[str]
     name: str
     container: str
-    pods: list[PodData]
     namespace: str
     kind: str
     allocations: ResourceAllocations
+    pods: list[PodData]
```
Avi-Robusta, line 24: do we still need the pods list anymore, can we get rid of all of that code?

Avi-Robusta: nvrmnd I see we still want to send it back

LeaveMyYard:
Closing as different approach to optimization was chosen

PR #89, opened 2023-06-28
fixing threading update bug

Avi-Robusta:


Avi-Robusta:
redundant, already fixed

PR #92, opened 2023-07-02
Fixed prometheus query memory issue

Avi-Robusta:
added override memory query for simple strategy so will not effect custom strategies
fixed bug creating fake memory spikes when a container restarts


pablos44 (approved):
This can work as a quick solution for resolving Prometheus memory consumption issues while keeping backward compatibility.

PR #93, opened 2023-07-03
Support prom discovery

### LINKED ISSUE #59: Support metrics-based workload discovery
**Is your feature request related to a problem? Please describe.**

In some cases, we want to run the KRR program locally. But for the security consideration, the API server of the Kubernetes cluster cannot be accessed outside of the cluster.

So we can use the Prometheus-based workload discovery if `kube-state-metrics` is installed.

**Describe the solution you'd like**

We can do the workload based discovery with the following steps,

1. List Deployments together with their ReplicaSets, 

```python
replicasets = await self.metrics_loader.loader.query("count by (namespace, owner_name, replicaset) (kube_replicaset_owner{"
                                               f'namespace=~"{ns}", '
                                               'owner_kind="Deployment"})')
 ```

2. List Pods from a group of ReplicaSets

```python
# owner_name is ReplicaSet names
pods = await self.metrics_loader.loader.query("count by (owner_name, replicaset, pod) (kube_pod_owner{"
                                               f'namespace="{namespace}", '
                                               f'owner_name=~"{owner_name}", '
                                               'owner_kind="ReplicaSet"})')
```

3. List containers from Pods got from step (2)

```python
containers = await self.metrics_loader.loader.query("count by (container) (kube_pod_container_info{"
                                               f'namespace="{namespace}", '
                                               f'pod=~"{pod_selector}"'
                                               "})")
```

4. Build K8sObjectData for containers got from step (3)

```python
async def __build_from_owner(self, namespace: str, app_name: str, containers: List[str], pod_names: List[str]) -> List[K8sObjectData]:
        return [
            K8sObjectData(
                cluster=None,
                namespace=namespace,
                name=app_name,
                kind="Deployment",
                container=container_name,
                allocations=await self.__parse_allocation(namespace, "|".join(pod_names), container_name), # find 
                pods=[PodData(name=pod_name, deleted=False) for pod_name in pod_names], # list pods
            )
            for container_name in containers
        ]
```

**Describe alternatives you've considered**

**Additional context**

lujiajing1126:
Closes #59 

According to the discussion in the thread, an extra option is added, 

```
--discovery-method           TEXT  Method to discover workload in the cluster. [default: api-server]
```

The default mode is `api-server` but can be switched to `prometheus`.

Currently, the following entity is supported,

 * [x] Deployment
 * [ ] STS
 * [ ] ...

### DIFF
--- a/robusta_krr/core/models/config.py
+++ b/robusta_krr/core/models/config.py
@@ -16,7 +16,7 @@ class Config(pd.BaseSettings):
     clusters: Union[list[str], Literal["*"], None] = None
     kubeconfig: Optional[str] = None
     namespaces: Union[list[str], Literal["*"]] = pd.Field("*")
-    discovery_method: Union[Literal["api-server"], Literal["prometheus"]] = pd.Field("api-server")
+    discovery_method: Literal["api-server", "prometheus"] = pd.Field("api-server")
 
     # Value settings
     cpu_min_value: int = pd.Field(5, ge=0)  # in millicores
--- a/robusta_krr/core/models/config.py
+++ b/robusta_krr/core/models/config.py
@@ -16,6 +16,7 @@ class Config(pd.BaseSettings):
     clusters: Union[list[str], Literal["*"], None] = None
     kubeconfig: Optional[str] = None
     namespaces: Union[list[str], Literal["*"]] = pd.Field("*")
+    discovery_method: Union[Literal["api-server"], Literal["prometheus"]] = pd.Field("api-server")
 
     # Value settings
     cpu_min_value: int = pd.Field(5, ge=0)  # in millicores
--- a/robusta_krr/core/models/config.py
+++ b/robusta_krr/core/models/config.py
@@ -17,6 +17,7 @@ class Config(pd.BaseSettings):
     kubeconfig: Optional[str] = None
     namespaces: Union[list[str], Literal["*"]] = pd.Field("*")
     discovery_method: Literal["api-server", "prometheus"] = pd.Field("api-server")
+    selector: Optional[str] = None
 
     # Value settings
     cpu_min_value: int = pd.Field(5, ge=0)  # in millicores
--- a/robusta_krr/core/models/config.py
+++ b/robusta_krr/core/models/config.py
@@ -26,9 +26,10 @@ class Config(pd.BaseSettings):
     # Prometheus Settings
     prometheus_url: Optional[str] = pd.Field(None)
     prometheus_auth_header: Optional[str] = pd.Field(None)
+    prometheus_other_headers: dict[str, str] = pd.Field(default_factory=dict)
     prometheus_ssl_enabled: bool = pd.Field(False)
     prometheus_cluster_label: Optional[str] = pd.Field(None)
-    prometheus_label: str = pd.Field("cluster")
+    prometheus_label: Optional[str] = pd.Field(None)
 
     # Threading settings
     max_workers: int = pd.Field(6, ge=1)
@@ -51,6 +52,13 @@ class Config(pd.BaseSettings):
     @property
     def Formatter(self) -> formatters.FormatterFunc:
         return formatters.find(self.format)
+
+    @pd.validator("prometheus_other_headers", pre=True)
+    def validate_prometheus_other_headers(cls, headers: Union[list[str], dict[str, str]]) -> dict[str, str]:
+        if isinstance(headers, dict):
+            return headers
+
+        return {k.strip().lower(): v.strip() for k, v in [header.split(":") for header in headers]}
 
     @pd.validator("namespaces")
     def validate_namespaces(cls, v: Union[list[str], Literal["*"]]) -> Union[list[str], Literal["*"]]:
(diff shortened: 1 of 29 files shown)


robusta_krr/core/runner.py
```diff
@@ -22,13 +23,23 @@
 from robusta_krr.utils.progress_bar import ProgressBar
 from robusta_krr.utils.version import get_version
 
+async def gather_with_concurrency(n: int, *coros):
```
lujiajing1126, line 26: https://github.com/robusta-dev/krr/pull/82 is not enough. Data race may happen if there are hundreds of deployment.

LeaveMyYard: Can you show/explain me when can it fail? Executor can only process max_workers amount of tasks at a time and we were worried exactly about limiting the amount of requests sent to prometheus.

What data race exactly we should be aware of? As I am aware, we do not modify state the way to get a data race.

robusta_krr/core/models/config.py
```diff
@@ -16,6 +16,7 @@ class Config(pd.BaseSettings):
     clusters: Union[list[str], Literal["*"], None] = None
     kubeconfig: Optional[str] = None
     namespaces: Union[list[str], Literal["*"]] = pd.Field("*")
+    discovery_method: Union[Literal["api-server"], Literal["prometheus"]] = pd.Field("api-server")
```
LeaveMyYard, line 19: You should be able to do like that:

`Literal["api-server", "prometheus"]`

robusta_krr/core/runner.py
lujiajing1126: > Can you show/explain me when can it fail? Executor can only process max_workers amount of tasks at a time and we were worried exactly about limiting the amount of requests sent to prometheus.
> 
> What data race exactly we should be aware of? As I am aware, we do not modify state the way to get a data race.

If we have many deployments, let's say 1000. Then 1000 coroutines (aka. tasks) will be created concurrently to gather data.

https://github.com/robusta-dev/krr/blob/9cdc002b9f4a754976dafe5a0452b3ac0242a669/robusta_krr/core/runner.py#L140

Since the max number of workers has been set to a fixed number, for example 6 by default, multiple promQL requests will be queued up. Speically there would be 1000 promQL requested created by `add_history_pods` from different coroutines,

https://github.com/robusta-dev/krr/blob/9cdc002b9f4a754976dafe5a0452b3ac0242a669/robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py#L156

So the second step of a single task cannot be started until all `add_history_pods` are finished. Then all tasks must hold their data in the memory. (I'm not an expert of Python. Pls correct me if I said something wrong)

From my observation, the memory of the process was increasing (> 2GB) while the progress bar was still.

robusta_krr/core/models/config.py
lujiajing1126: Fixed

robusta_krr/core/runner.py
LeaveMyYard: Yeah, I think you are right
But let's create a separate issue and I will try to fix it in a separate PR

LeaveMyYard: Want to take a deeper look into this one
So please remove the `gather_with_concurrency` for now

lujiajing1126: > Want to take a deeper look into this one So please remove the `gather_with_concurrency` for now

Removed now

LeaveMyYard:
Planned to merge in 1.7

danielhass:
@LeaveMyYard this apparently didn't make it into 1.7. Can you guys share the current plans for this PR?

aantn:
We're looking into this! It needs some updating for the current codebase and is currently incomplete.

So we can't commit to a timeline yet, but it is something we want to merge in principle.

There are two scenarios we're familiar with where this would be useful:

1. For people who don't have kubectl access to the cluster
2. For ephemeral workloads that no longer exist at the time of the scan (e.g. gitlab runners)

Just to confirm we're on the right track with this, is your use case one of those two?

aantn:
And if there are any volunteers for updating this PR to the current version of Robusta and completing it - that would definitely accelerate how fast we can merge it.

danielhass:
@aantn thanks for providing this feedback here!

Our use case falls into these cases and extends beyond I would say: we too don't want to require kubectl to the scanned cluster to run krr. Additionally in our landscape we consolidate our metrics on multiple central "infrastructure clusters" via Prom remote-writes. These infra clusters also store metrics with a much larger retention than the local Prom instances in the source/workload clusters. Running krr on these infra clusters would allow us to fit the tool much better in the centralized metrics approach we chose and possibly lead to better results as we have much more data (larger metrics retention) available there.

Does this makes sense?

aantn:
Yep, makes sense.

You may be able to do that today without waiting for the PR. You can pass an explicit Prometheus url with `-p` and use `--prometheus-label` and `-l` to specify which metrics belong to each cluster. See "Scanning with a Centralized Prometheus" in [this part of the README](https://github.com/robusta-dev/krr?tab=readme-ov-file#data-source-integrations).

Let me know if that works for your case?

danielhass:
@aantn we could reproduce your edited findings 💯 on our side.

We were indeed able to make use of the lager retention on the central Prom instance with mentioned parameters. However krr failed as soon as we removed the kubectl access to the target cluster as this connection was always used for workload discovery.

aantn:
Thank you for the update! That makes sense.

PR #94, opened 2023-07-03
Add `executor` init arg to MetricsService

hateyouinfinity:
Fixes a bug likely introduced in #82.
`MetricsLoader.get_metrics_service` passes executor to `VictoriaMetricsService` and `ThanosMetricsService`, but those classes don't have a corresponding init arg, so an exception is always thrown if loop reaches them.


Avi-Robusta:
hey, good catch, we already merged a fix the other day but thank you for opening the PR

PR #97, opened 2023-07-05
Prometheus post usage for queries

### LINKED ISSUE #60: Too many historic pods causes querystring too long for Prometheus range_query API

LeaveMyYard:
Make queries to Prometheus use POST API, as it has bigger limit then query params

Will fix #60 and #66, maybe #62 also

### DIFF
--- a/robusta_krr/core/integrations/prometheus/prometheus_client.py
+++ b/robusta_krr/core/integrations/prometheus/prometheus_client.py
@@ -1,0 +1,89 @@
+from typing import no_type_check
+
+import requests
+from datetime import datetime
+from prometheus_api_client import PrometheusConnect, Retry, PrometheusApiClientException
+from requests.adapters import HTTPAdapter
+
+
+class ClusterNotSpecifiedException(Exception):
+    """
+    An exception raised when a prometheus requires a cluster label but an invalid one is provided.
+    """
+
+    pass
+
+
+class CustomPrometheusConnect(PrometheusConnect):
+    """
+    Custom PrometheusConnect class to handle retries.
+    """
+
+    def __init__(
+        self,
+        url: str = "http://127.0.0.1:9090",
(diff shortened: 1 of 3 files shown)


robusta_krr/core/integrations/prometheus/prometheus_client.py
```diff
@@ -0,0 +1,88 @@
...
+        return data
+
+    def custom_query_range(
+        self,
+        query: str,
+        start_time: datetime,
+        end_time: datetime,
+        step: str,
+        params: dict = None,
+    ):
```
Avi-Robusta, line 62: adding this could solve future errors
```
        starts_at = starts_at.replace(microsecond=0)
        ends_at = ends_at.replace(microsecond=0)
```
There are some versions of prometheus that return empty queries when passed microseconds
like GKE managed prometheus

Avi-Robusta (approved):
looks good, one potential thing to add if you have time

robusta_krr/core/integrations/prometheus/prometheus_client.py
LeaveMyYard: Rounding should do the same job (see below)

PR #99, opened 2023-07-06
Fallback to HPAv1 if HPAv2 not available

LeaveMyYard:
Some older versions of kubernetes does not have HPAv2 API available.

Use HPAv1 API for listing HPA if v2 is not available

### DIFF
--- a/robusta_krr/core/integrations/kubernetes.py
+++ b/robusta_krr/core/integrations/kubernetes.py
@@ -29,6 +29,7 @@ from robusta_krr.utils.configurable import Configurable
 
 
 AnyKubernetesAPIObject = Union[V1Deployment, V1DaemonSet, V1StatefulSet, V1Pod, V1Job]
+HPAKey = tuple[str, str, str]
 
 
 class ClusterLoader(Configurable):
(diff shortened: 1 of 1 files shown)


robusta_krr/core/integrations/kubernetes.py
```diff
@@ -216,9 +238,20 @@ async def __list_hpa(self) -> dict[tuple[str, str], HPAData]:
         """
 
         loop = asyncio.get_running_loop()
-        res: V2HorizontalPodAutoscalerList = await loop.run_in_executor(
-            self.executor, lambda: self.autoscaling.list_horizontal_pod_autoscaler_for_all_namespaces(watch=False)
-        )
+
+        try:
+            # Try to use V2 API first
+            res: V2HorizontalPodAutoscalerList = await loop.run_in_executor(
```
Avi-Robusta, line 244: do we need to return the result?

Avi-Robusta: do we have to format the result to be of object HPAData

Avi-Robusta (approved):
hey looks good, had a few questions
also wanted to ask if you ran both functions by commenting out some lines just to see they both work?

Avi-Robusta:
also dont forget to rebase

robusta_krr/core/integrations/kubernetes.py
LeaveMyYard: It is formatted below. 
`__get_metric` is the inner function

PR #116, opened 2023-07-24
Refactor strategy dependancy on metrics

LeaveMyYard:
Now strategy defines itself the list of required metrics. Metrics are now loaded dynamically based on that.


robusta_krr/core/integrations/prometheus/metrics/cpu.py
```diff
@@ -0,0 +1,33 @@
...
+    def get_query(self, object: K8sObjectData, resolution: Optional[str]) -> str:
+        pods_selector = "|".join(pod.name for pod in object.pods)
+        cluster_label = self.get_prometheus_cluster_label()
+        return (
+            "sum(irate(container_cpu_usage_seconds_total{"
+            f'namespace="{object.namespace}", '
+            f'pod=~"{pods_selector}", '
+            f'container="{object.container}"'
+            f"{cluster_label}"
+            "}[5m])) by (container, pod, job)"
```
Avi-Robusta, line 32: this looks incorrect there is no max here and the resolution parameter isn't used so it wont take into account the last two weeks

robusta_krr/utils/progress_bar.py
```diff
@@ -14,7 +14,7 @@ class ProgressBar(Configurable):
 
     def __init__(self, config: Config, **kwargs) -> None:
         super().__init__(config)
-        self.show_bar = self.echo_active
+        self.show_bar = self.echo_active and False
```
Avi-Robusta, line 17: wont this always hide the bar?

robusta_krr/core/abstract/strategies.py
```diff
@@ -61,21 +64,9 @@ def timeframe_timedelta(self) -> datetime.timedelta:
...
-    data is a mapping from pod to a numpy array of time and value.
-    """
-
-    metric: Metric
-    data: dict[str, ArrayNx2]  # Mapping: pod -> [(time, value)]
-
-    class Config:
-        arbitrary_types_allowed = True
+MetricPodData = dict[str, ArrayNx2]  # Mapping: pod -> [(time, value)]
+MetricsPodData = dict[type["BaseMetric"], MetricPodData]
```
Avi-Robusta, line 68: MetricsPodData and MetricPodData are sooooo simmilar, it confused me for a bit I would recommend renaming one of them to prevent future mixups.

robusta_krr/core/models/result.py
```diff
@@ -22,26 +21,13 @@ class ResourceRecommendation(pd.BaseModel):
...
-
-
-MetricsData = dict[ResourceType, Metric]
-
-
 class ResourceScan(pd.BaseModel):
     object: K8sObjectData
     recommended: ResourceRecommendation
     severity: Severity
-    metrics: MetricsData
```
Avi-Robusta, line 39: i think this changes were needed to display the queries used in the json output so that we could eventually make graphs in frontend.
I dont see the queries now when i run it with -f json

Avi-Robusta (changes requested):
looks good,
The query needs to be fixed for max cpu 
and we don't get the graphs in the output json format

robusta_krr/core/models/result.py
LeaveMyYard: We discussed that with @pablos44 and @yuryrudey 
Frontend will use it's metrics

Avi-Robusta (approved):

robusta_krr/core/integrations/prometheus/metrics/cpu.py
```diff
@@ -0,0 +1,39 @@
...
+                )
+            ) by (container, pod, job)
+        """
+
+
+class MaxCPULoader(QueryMetric, FilterMetric):
+    def get_query(self, object: K8sObjectData, resolution: str) -> str:
+        pods_selector = "|".join(pod.name for pod in object.pods)
+        cluster_label = self.get_prometheus_cluster_label()
+        return f"""
```
Avi-Robusta, line 28: not doing max here

Avi-Robusta: also check the output makes sense

PR #118, opened 2023-07-26
 Optimization on memory usage

LeaveMyYard:


robusta_krr/core/integrations/kubernetes.py
```diff
@@ -63,32 +63,31 @@ async def list_scannable_objects(self) -> list[K8sObjectData]:
...
+            self._list_all_jobs(),
+        ]
+
+        for fut in asyncio.as_completed(tasks):
+            try:
+                object_list = await fut
+            except Exception as e:
+                self.error(f"Error {e.__class__.__name__} listing objects in cluster {self.cluster}: {e}")
+                self.debug_exception()
+                self.error("Will skip this object type and continue.")
```
Avi-Robusta, line 82: did people have issues with us not being able to get a specific object type?
(rollouts?)

robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py
```diff
@@ -186,25 +186,47 @@ async def add_historic_pods(self, object: K8sObjectData, period: datetime.timede
...
+                    }}[{period_literal}]
+                )
+            """
+        )
+
+        if related_pods == []:
+            self.debug(f"No pods found for {object}")
+            return
+
+        current_pods = await self.query(
```
Avi-Robusta, line 213: Did you do this for memory purposes or for accuracy? (i like it)

robusta_krr/core/abstract/strategies.py
```diff
@@ -98,20 +88,25 @@ class BaseStrategy(abc.ABC, Generic[_StrategySettings]):
...
     The strategy will automatically be registered in the strategy registry using __subclasses__ mechanism.
     """
 
-    __display_name__: str
-
-    settings: _StrategySettings
+    display_name: str
+    rich_console: bool = False
+    # TODO: this should be BaseMetricLoader, but currently we only support Prometheus
+    metrics: list[type[PrometheusMetric]] = []
```
Avi-Robusta, line 99: not sure I get this?

robusta_krr/core/integrations/kubernetes.py
LeaveMyYard: Yes, that's why I descided that soft error is better, so people will still get some data

LeaveMyYard: The rollout issue was fixed tho

robusta_krr/core/abstract/strategies.py
LeaveMyYard: This is from the previous PR

robusta_krr/core/runner.py
```diff
@@ -143,29 +143,27 @@ async def _calculate_object_recommendations(self, object: K8sObjectData) -> RunR
...
-                    requests={resource: recommendation[resource].request for resource in ResourceType},
-                    limits={resource: recommendation[resource].limit for resource in ResourceType},
-                    info={resource: recommendation[resource].info for resource in ResourceType},
-                )
-            )
-            for recommendation in recommendations
-        ]
+        self.__progressbar.progress()
+
+        return ResourceScan.calculate(
```
Avi-Robusta, line 159: out of curiosity does calculating here improve krr memory?

robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py
```diff
@@ -183,25 +186,47 @@ async def add_historic_pods(self, object: K8sObjectData, period: datetime.timede
...
-            "}"
-            f"[{period_literal}]"
+            f"""
+                last_over_time(
+                    kube_pod_owner{{
+                        owner_name=~"{owners_regex}",
+                        owner_kind="{pod_owner_kind}",
+                        namespace="{object.namespace}"
+                        {cluster_label}
+                    }}[{period_literal}]
```
Avi-Robusta, line 204: do we need anything from this other then the pod names?
we could change the query to only return one label to lower memory
(same with below)

Avi-Robusta (approved):
looks good, 
If we only get the pod label it might lower the memory a little bit too (but probably not much)

robusta_krr/core/runner.py
LeaveMyYard: This is for a refactoring

LeaveMyYard: This is only for a refactoring

robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py
LeaveMyYard: Here we actually need to get last time point for each, not all.
But with labels it will not make that much difference I think, but we can do it in future

PR #120, opened 2023-08-02
Implement prometheus request batching

LeaveMyYard:
Added batching for prometheus requests, if too many pods are present


robusta_krr/strategies/simple.py
```diff
@@ -38,7 +36,7 @@ def calculate_cpu_proposal(self, data: PodsTimeData) -> float:
         else:
             data_ = list(data.values())[0][:, 1]
 
-        return np.percentile(data_, self.cpu_percentile)
+        return np.max(data_)
```
Avi-Robusta, line 39: why are we switching from percentile to max here?

Avi-Robusta (approved):

robusta_krr/strategies/simple.py
LeaveMyYard: Because CPU is used in the PromQL here

robusta_krr/core/integrations/prometheus/metrics/base.py
```diff
@@ -228,3 +229,68 @@ def filter_prom_jobs_results(
     async def query_prometheus(self, data: PrometheusMetricData) -> list[PrometheusSeries]:
         result = await super().query_prometheus(data)
         return self.filter_prom_jobs_results(result)
+
+
+class BatchedRequestMixin(PrometheusMetric):
```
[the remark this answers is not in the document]
LeaveMyYard: Here we now use the maximum of percentiles for CPU

PR #48, opened 2023-05-25
Support filtering by resource type, and more focused namespace retrieval. 

mnbbrown:
This allows non-admin users on RBAC enforced clusters to run krr.

It does this by:

- Adding a new CLI flag called "resources" which allows users to only
  run krr on specific resources.

- Using the `_namespaced_` functions in the k8s python client library
  where specific namespaces are being queried.

PR #122, opened 2023-08-04
Fix listing namespaced workflows

LeaveMyYard:
Similar to #48, This allows non-admin users on RBAC enforced clusters to run krr.

This allows non-admin users on RBAC enforced clusters to run krr.

It does this by:
- Adding a new CLI flag called "resources" which allows users to only
run krr on specific resources.
- Using  namespaced functions in the k8s python client library
where specific namespaces are being queried.

Thanks to @mnbbrown for initial PR


robusta_krr/core/integrations/rollout.py
```diff
@@ -1,10 +1,7 @@
 from kubernetes import client
```
Avi-Robusta, line 1: is this from a specific version of the kubernetes cli?
Any reason why we dont just upgrade the kubernetes cli to the relevant version instead of having this file?

robusta_krr/core/models/config.py
```diff
@@ -68,6 +69,13 @@ def validate_namespaces(cls, v: Union[list[str], Literal["*"]]) -> Union[list[st
         if v == []:
             return "*"
 
+        return [val.lower() for val in v]
+
+    @pd.validator("resources")
+    def validate_resources(cls, v: Union[list[str], Literal["*"]]) -> Union[list[str], Literal["*"]]:
+        if v == []:
+            return "*"
+
```
Avi-Robusta, line 78: do we validate that all the resources are ones we support? if someone mispells deployment what will happen?

Avi-Robusta (approved):
looks good,
A few questions.
add small change to resource validator

PR #128, opened 2023-08-18
Calculation improvement

LeaveMyYard:
- Increased memory buffer
- Changed default step to 1m15s (1.25m)
- Fixed duration and step formatting in queries

### DIFF
--- a/robusta_krr/strategies/simple.py
+++ b/robusta_krr/strategies/simple.py
@@ -43,6 +43,8 @@ class SimpleStrategy(BaseStrategy[SimpleStrategySettings]):
     """
     CPU request: {cpu_percentile}% percentile, limit: unset
     Memory request: max + {memory_buffer_percentage}%, limit: max + {memory_buffer_percentage}%
+    History: {history_duration} hours
+    Step: {timeframe_duration} minutes
 
     This strategy does not work with objects with HPA defined (Horizontal Pod Autoscaler).
     If HPA is defined for CPU or Memory, the strategy will return "?" for that resource.
--- a/robusta_krr/strategies/simple.py
+++ b/robusta_krr/strategies/simple.py
@@ -11,13 +11,22 @@ from robusta_krr.core.abstract.strategies import (
     RunResult,
     StrategySettings,
 )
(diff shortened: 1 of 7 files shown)


fcecagno:
Just tested this and found it odd that all my deployments had recommendation of 100m CPU request, even those that consumed almost no CPU in the past (for instance, `config-reloader`).

LeaveMyYard:
@fcecagno I think this is because we set the minimal recommendation to 100m for CPU (previously it was 5m I think). @Avi-Robusta Don't you think that we should make it less?

robusta_krr/strategies/simple.py
```diff
@@ -75,14 +94,22 @@ def __calculate_memory_proposal(
         self, history_data: MetricsPodData, object_data: K8sObjectData
     ) -> ResourceRecommendation:
         data = history_data["MaxMemoryLoader"]
+        data_count = {pod: values[0, 1] for pod, values in history_data["MemoryAmountLoader"].items()}
```
RoiGlinik, line 97: this can be done after checking if data is empty

robusta_krr/strategies/simple.py
```diff
@@ -55,16 +66,24 @@ class SimpleStrategy(BaseStrategy[SimpleStrategySettings]):
...
     @property
     def metrics(self) -> list[type[PrometheusMetric]]:
-        return [PercentileCPULoader(self.settings.cpu_percentile), MaxMemoryLoader]
+        return [PercentileCPULoader(self.settings.cpu_percentile), MaxMemoryLoader, CPUAmountLoader, MemoryAmountLoader]
 
     def __calculate_cpu_proposal(
         self, history_data: MetricsPodData, object_data: K8sObjectData
     ) -> ResourceRecommendation:
         data = history_data["PercentileCPULoader"]
+        data_count = {pod: values[0, 1] for pod, values in history_data["CPUAmountLoader"].items()}
```
RoiGlinik, line 75: can be done after checking if data is empty

RoiGlinik (approved):

PR #131, opened 2023-08-22
Fix listing rollout pods, fix rollout type

LeaveMyYard:
Fix rollouts pod listing from prometheus
Fix bug, where rollouts were previously displayed as deployments

### DIFF
--- a/robusta_krr/core/integrations/kubernetes.py
+++ b/robusta_krr/core/integrations/kubernetes.py
@@ -129,7 +129,7 @@ class ClusterLoader(Configurable):
             self.debug(f"Skipping {kind}s in {self.cluster}")
             return
 
-        if kind == "rollout" and not self.__rollouts_available:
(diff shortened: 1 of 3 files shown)


robusta_krr/core/integrations/kubernetes.py
```diff
@@ -129,7 +129,7 @@ async def _list_workflows(
             self.debug(f"Skipping {kind}s in {self.cluster}")
             return
 
-        if kind == "rollout" and not self.__rollouts_available:
+        if kind == "Rollout" and not self.__rollouts_available:
```
Avi-Robusta, line 132: any reason why you changed the casing?

Avi-Robusta (approved):

robusta_krr/core/integrations/kubernetes.py
LeaveMyYard: To be used in object kind (previously they were taken from API response)

PR #140, opened 2023-09-04
Update simple.py

reAsOn2010:
Simple strategy cpu calculation differs from memory calculation. I thing cpu calculation should also use the data filtered.


LeaveMyYard (approved):
LGTM! Sign the CLA and I'll merge it.

PR #141, opened 2023-09-04
Refactor mixins for prometheus metrics

LeaveMyYard:
Rework usage of mixin pattern to make code more readable

### DIFF
--- a/robusta_krr/core/integrations/prometheus/metrics/base.py
+++ b/robusta_krr/core/integrations/prometheus/metrics/base.py
@@ -2,12 +2,11 @@ from __future__ import annotations
 
 import abc
 import asyncio
-import copy
 import datetime
 import enum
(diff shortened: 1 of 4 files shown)


robusta_krr/core/integrations/prometheus/metrics/base.py
```diff
@@ -55,6 +74,9 @@ def __init__(
 
         self.executor = executor
 
+        if self.pods_batch_size is not None and self.pods_batch_size <= 0:
+            raise ValueError("pods_batch_size must be positive")
```
Avi-Robusta, line 78: how could this be negative?

Avi-Robusta (approved):
looks good,
Just the one question about batches.
Did you check the output to verify its the same?

LeaveMyYard:
> Did you check the output to verify its the same?
Results are exactly the same as before changes

Issue #127, opened 2023-08-16
Krr 1.5.3 returns no results due to no metrics for PercentileCPULoader and MaxMemoryLoader


agabrys:
**Describe the bug**

I updated the tool to the [1.5.3](https://github.com/robusta-dev/krr/releases/tag/v1.5.3) version. When I executed the `simple` strategy and it returned no results:

```
> krr simple --namespace <namespace> --selector="app = grafana"


 _____       _               _          _  _______  _____
|  __ \     | |             | |        | |/ /  __ \|  __ \
| |__) |___ | |__  _   _ ___| |_ __ _  | ' /| |__) | |__) |
|  _  // _ \| '_ \| | | / __| __/ _` | |  < |  _  /|  _  /
| | \ \ (_) | |_) | |_| \__ \ || (_| | | . \| | \ \| | \ \
|_|  \_\___/|_.__/ \__,_|___/\__\__,_| |_|\_\_|  \_\_|  \_\



Running Robusta's KRR (Kubernetes Resource Recommender) v1.5.3
Using strategy: Simple
Using formatter: table

[INFO] Using clusters: ['<cluster>']
on 0: [INFO] Listing scannable objects in <cluster>
on 0: [INFO] Connecting to Prometheus for <cluster> cluster
on 0: [INFO] Using Prometheus at https://<server>/api/v1/namespaces/<namespace>/services/prometheus-server-service:9091/proxy for cluster <cluster>   
on 0: [INFO] Prometheus found
on 0: [INFO] Prometheus connected successfully for <cluster> cluster
on 0: [WARNING] Prometheus returned no PercentileCPULoader metrics for Deployment <namespace>/grafana-deployment/grafana
on 0: [WARNING] Prometheus returned no MaxMemoryLoader metrics for Deployment <namespace>/grafana-deployment/grafana
Calculating Recommendation |████████████████████████████████████████| 1 in 15.4s (0.07/s)



Simple Strategy

CPU request: 99.0% percentile, limit: unset
Memory request: max + 5.0%, limit: max + 5.0%

This strategy does not work with objects with HPA defined (Horizontal Pod Autoscaler).
If HPA is defined for CPU or Memory, the strategy will return "?" for that resource.

Learn more: https://github.com/robusta-dev/krr#algorithm

[...]

PR #150, opened 2023-10-06
Fallback to K8S API for getting pods, refactor logging for new warnings

LeaveMyYard:
Fallback to loading pods from KubeAPI in case some error occured with prometheus.
Raise a warning, as this results in missing some pods from the total calculation.

Note: previously (<= 1.4) KRR was loading pods from the API, and then loading deleted ones from prometheus (as only prometheus keeps track of historical objects). 1.5 introduced an optimization of loading everything only from prometheus, but some users (https://github.com/robusta-dev/krr/issues/127#issuecomment-1737223230) seem to not have some required metrics.

Added a section in a readme, describing what metrics are required to be present on a Prometheus, add a link to that in a CLI warning


README.md
```diff
@@ -87,6 +87,20 @@ Read more about [how KRR works](#how-it-works) and [KRR vs Kubernetes VPA](#diff
 
 ## Installation
 
+### Requirements
+
```
aantn, line 91: I think you can remove `running either on your cluster or centralized.`

We use the term "centralized" a lot, but not sure its obvious to the user what we mean.

README.md
```diff
@@ -87,6 +87,20 @@ Read more about [how KRR works](#how-it-works) and [KRR vs Kubernetes VPA](#diff
...
+KRR requires you to have Prometheus, running either on your cluster or centralized.
+
+Additionally to that, [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) needs to be running on your cluster, as KRR is dependant on next metrics:
```
aantn, line 94: `next` -> `those`

README.md
```diff
@@ -87,6 +87,20 @@ Read more about [how KRR works](#how-it-works) and [KRR vs Kubernetes VPA](#diff
...
+
+- `container_cpu_usage_seconds_total`
+- `container_memory_working_set_bytes`
+- `kube_replicaset_owner`
+- `kube_pod_owner`
+- `kube_pod_status_phase`
+
+_Note: If one of last three metrics is absent KRR will still work, but will result in potentially worse results and raise a warning_
```
aantn, line 102: Can you explain why they will be worse?

LeaveMyYard: It is explained in a PR description:

>Fallback to loading pods from KubeAPI in case some error occured with prometheus.
Raise a warning, as this results in missing some pods from the total calculation.

>Note: previously (<= 1.4) KRR was loading pods from the API, and then loading deleted ones from prometheus (as only prometheus keeps track of historical objects). 1.5 introduced an optimization of loading everything only from prometheus, but some users (https://github.com/robusta-dev/krr/issues/127#issuecomment-1737223230) seem to not have some required metrics.

aantn: Yes, but it will be missing for future readers of the README who don't know anything about this PR.

Assuming I understood correctly, I suggest:

_Note: If one of last three metrics is absent KRR will still work, but it will only consider currently-running pods when calculating recommendations. Historic pods that no longer exist in the cluster will not be taken into consideration._

arikalon1 (approved):
LGTM

Please fix Natan's comments on the documentation and merge

PR #155, opened 2023-10-10
Patch fix for proxy not working from kubeconfig

### LINKED ISSUE #44: Does not work with a proxy
I need a proxy to get to our kubernetes environments. 
The proxy is specified in kubeconfig but it seems krr aren't using it.
I also tried to set an http{s}_proxy env variable but that didn't work either.

Using krr v1.0.0.

LeaveMyYard:
fixes #44


arikalon1 (approved):
LGTM

Issue #158, opened 2023-10-13
How to connect with mimir ?


pingping95:
**Is your feature request related to a problem? Please describe.**

krr simple -p https://mimir.xxxxx.xxx --prometheus-label cluster -l my-cluster

I tried above command and does not success.

<img width="760" alt="image" src="https://github.com/robusta-dev/krr/assets/67780144/db91b41d-fcb8-400d-8b04-ff33f45fe4ba">

**Describe the solution you'd like**
Hope to support Mimir

If i did wrong command, please let me know the exact command.

KelvinVenancio:
I am looking forward to being able to use it with Mimir too.

FrankFoerster24:
Hi,

this works for me:
```bash
krr simple --prometheus-url "http://localhost:8080/prometheus"
```

This is a Mimir deployment using the `mimir-distributed-4.4.1` helm chart, `:8080` is a port forward to its Nginx service (or rather a pod behind it), the /prometheus route should hit Mimir's query frontend.  So any URL that hits this service should work, I just happened to use port forwarding.  Note that it is a bit sensitive to trailing slashes, eg. I noticed `http://localhost:8080/prometheus/` does not work for me.  Good luck.

Kind regards,
 Frank.

KelvinVenancio:
Hi @FrankFoerster24, perfect, it works like a charm, thank you for sharing with us.

aantn:
Thanks for the update.

If anyone would like to contribute a PR to fix the docs (or autodiscovery, or that issue with trailing slashes) that would be more than welcome! And we'll send you a cool Robusta t-shirt in appreciation :)

FrankFoerster24:
As a starting point, I hacked something in #168 which works for my installation.  Some things are not at the right place yet, though.

nantiferov:
I can add to aforementioned example that current version of krr (v1.6.0) works without issues with self-hosted Mimir.

Only headers for tenant-id should be provided too (were added in https://github.com/robusta-dev/krr/issues/84), example

```shell
$ krr simple --prometheus-url https://mimir.some.domain/prometheus --prometheus-headers 'X-Scope-OrgID: some-tenant-id'
```

aantn:
@pavangudiwada could you assist with updating the docs re. @FrankFoerster24's and @nantiferov's instructions?

pingping95:
Hi all.

I think cluster-label config need to add for those who have lots cluster label values for multi cluster.

```
[...]

PR #168, opened 2023-11-07
Add Mimir support

FrankFoerster24:


robusta_krr/core/integrations/prometheus/metrics_service/mimir_metrics_service.py
FrankFoerster24, line 1: This is pretty much a copy of `victoria_metrics_service.py` substituting `[Vv]ictoria` with `[Mm]imir`.

robusta_krr/core/integrations/prometheus/metrics_service/mimir_metrics_service.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Checks the connection to Prometheus.
+        Raises:
+            MimirMetricsNotFound: If the connection to Mimir Metrics cannot be established.
+        """
+        try:
+            super().check_connection()
+        except MetricsNotFound as e:
+            # This is to clarify which metrics service had the issue and not say its a prometheus issue
+            raise MetricsNotFound(
```
FrankFoerster24, line 43: AFAICS `prometrix` has no specific `MimirMetricsNotFound`, so I'm just raising its base exception.

robusta_krr/core/integrations/prometheus/metrics_service/mimir_metrics_service.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Finds the Mimir Metrics URL using selectors.
+        Args:
+            api_client (Optional[ApiClient]): A Kubernetes API client. Defaults to None.
+        Returns:
+            Optional[str]: The discovered Mimir Metrics URL, or None if not found.
+        """
+        return super().find_url(
+            selectors=[
+                "app.kubernetes.io/name=mimir,app.kubernetes.io/component=query-frontend",
```
FrankFoerster24, line 21: That selector works for my installation and I don't know any others.  List can be extended, of course.

robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py
```diff
@@ -79,6 +79,10 @@ def __init__(
         logger.info(f"Using {self.name} at {self.url} for cluster {cluster or 'default'}")
 
         headers = settings.prometheus_other_headers
+        # FIXME: not the right place, Mimir specific
+        headers.update(
+            {"X-Scope-OrgID": "anonymous"}
```
FrankFoerster24, line 84: That's probably the wrong place, but I don't know where to put this.  It is also strange that it is needed: passing in a Mimir URL using `--prometheus-url` succeeds also without this header.  Also, this should be configurable: Mimir has multi-tenant capabilities and this header identifies the tenant (AFAIK).

robusta_krr/core/integrations/prometheus/prometheus_utils.py
```diff
@@ -25,9 +25,10 @@ def generate_prometheus_config(
     url: str, headers: dict[str, str], metrics_service: PrometheusMetricsService
 ) -> PrometheusConfig:
     from .metrics_service.victoria_metrics_service import VictoriaMetricsService
+    from .metrics_service.mimir_metrics_service import MimirMetricsService
 
     baseconfig = {
-        "url": url,
+        "url": url + "/prometheus" if isinstance(metrics_service, MimirMetricsService) else url,
```
FrankFoerster24, line 31: Also ugly, but it only works when adding this to the URL.

junze-smg: https://grafana.com/docs/mimir/latest/configure/configure-the-query-frontend-work-with-prometheus/
I think this can be avoided by setting `api.prometheus_http_prefix` to `''`.

robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py
junze-smg: L81 allows you to set other headers via command line arguments

ampc:
Any plans to support Mimir ?  we would like to try robusta krr, but we only use Mimir + grafana-agent stack.

FrankFoerster24:
> Any plans to support Mimir ?

@ampc, you can already use Mimir, see instructions in comments in the linked issue #158 .  Given that you can somehow access Mimir from the machine running krr.  This PR would just add autodiscovery for the Mimir query endpoint which would in turn enable in-cluster communication via the Kubernetes API.

LeaveMyYard (approved):
Thanks for your PR, finally we are merging it :)

PR #176, opened 2023-12-08
feat(VictoriaMetricsDiscovery): set correct url for cluster version

chicocvenancio:
Using victoriametrics cluster version will fail due to it needing a different url.


LeaveMyYard (approved):

robusta_krr/core/integrations/prometheus/metrics_service/victoria_metrics_service.py
```diff
@@ -17,14 +17,21 @@ def find_metrics_url(self, *, api_client: Optional[ApiClient] = None) -> Optiona
...
             ]
         )
+        if url is None:
+            url = super().find_url(
+                selectors=[
+                    "app.kubernetes.io/name=vmselect",
+                    "app=vmselect",
+                ]
+            )
+            url = f"{url}/select/0/prometheus/"
```
LeaveMyYard, line 33: Here url could be `None`, but I'll fix it myself

LeaveMyYard (approved):

PR #185, opened 2023-12-22
Update README.md: differences table is broken

seanorama:
There is an extra line break in the "difference" table which is causing the last rows to not show in the table.


aantn:
Thanks, good catch. Would it be possible for you to sign the CLA?

pavangudiwada:
Closing this is because we fixed it #188

PR #196, opened 2024-01-18
Fix Typo

mrueg:
Stafeful with a single L, in the CLI help users might get confused otherwise.


aantn (approved):

aantn:
Nice catch, thank you!

PR #209, opened 2024-02-13
Add Openshift support with loading sa token

LeaveMyYard:
Add --openshift flag. If it is set, try to load a service account token for connecting to Prometheus


arikalon1 (approved):
LGTM

PR #211, opened 2024-02-14
info about excluded namespace

### LINKED ISSUE #208: skips kube-system namespace despite saying it scans all namespaces by default
**Describe the bug**

krr says in help that by default it runs on all namespaces:

```
[☸ lab:database] [⏱ 2s]
❯ ~ krr simple --help | grep namespace
│ --namespace          -n      TEXT  List of namespaces to run on. By default, will run on all namespaces. [default: None]
```

This is not true. Seems it by default skip kube-system namespace.

**To Reproduce**
Run `krr simple` and then run `krr simple --namespace kube-system`.
Second invocation will list resources that were absent in first invocation.

**Expected behavior**
Either help should be updated to state that kube-system is excluded by default or default behaviour should be changed to include kube-system namespace.

fenio:
fixes #208


aantn (approved):
Thank you!

PR #215, opened 2024-02-27
Upgraded prometrix.

ganeshrvel:


LeaveMyYard (changes requested):
There is one more `custom_query_range` change at `robusta_krr/core/integrations/prometheus/metrics_service/prometheus_metrics_service.py`, shouldn't it also be changed?

LeaveMyYard (approved):

PR #219, opened 2024-03-02
Simplify code for main and strategies

aantn:


robusta_krr/main.py
```diff
@@ -26,281 +25,272 @@
...
-                            f"--{field_name}",
-                            help=f"{field_meta.field_info.description}",
-                            rich_help_panel="Strategy Settings",
-                        ),
-                        annotation=__process_type(field_meta.type_),
-                    )
-                    for field_name, field_meta in strategy_type.get_settings_type().__fields__.items()
-                ]
-            )
+@app.command(name="simple", rich_help_panel="Strategies")
```
LeaveMyYard, line 46: The problem here is that now in order to create a new strategy it is required to run this 200 lines thing
The previous approach removed the need for such thing

What I am thinking about is to make the command `run` and to pass a name of the strategy as a first parameter. Then still we need to somehow load the strategy from the known subclasses

robusta_krr/main.py
```diff
@@ -26,281 +25,272 @@
...
+    file_output: Optional[str] = typer.Option(
+        None, "--fileoutput", help="Print the output to a file", rich_help_panel="Output Settings"
+    ),
+    slack_output: Optional[str] = typer.Option(
+        None,
+        "--slackoutput",
+        help="Send to output to a slack channel, must have SLACK_BOT_TOKEN",
+        rich_help_panel="Output Settings",
+    ),
+    cpu_min_value: int = __pydantic_field_to_typer_option(
```
LeaveMyYard, line 218: Also here, it is now required to add those fields here manually. For our own codebase it is ok, but it will be much more code for someone to implement a strategy himself

Note: there were some users in discussions that told that they are making their own strategies, so I feel like such a big added boilerplate is bad

LeaveMyYard (changes requested):
We had an easy ability to create custom strategies that were used by some users. I want to keep it as it was or, at least the same easy.

There is a file for example and it will not work at the moment:
`examples/custom_strategy.py`

robusta_krr/main.py
aantn: It is worth the increased simplicity.

robusta_krr/main.py
aantn: I understand, but we're going to optimize for simplicity and getting more PRs, even at the expense of introducing boilerplate.

aantn:
I'll update the docs regarding custom strategies and remove a reference to the old method that doesn't work.

Going forward, can we support custom strategies by having users import KRR as a library and pass their own strategy in, just like we do in main.py()?

LeaveMyYard:
> I'll update the docs regarding custom strategies and remove a reference to the old method that doesn't work.
> 
> Going forward, can we support custom strategies by having users import KRR as a library and pass their own strategy in, just like we do in main.py()?

We can, it is really easy to add at the moment

skylightxo:
> # add a new command - most of this is boilerplate

What is the purpose of this change? We have a few custom strategies that we use inside CI pipelines and after this PR it will be much less convenient to write while previously everything was auto-discovered and just worked.

skylightxo:
Also if some new parameters will appear in new versions, we will always have errors as we will not have them automatically added to all strategies. So we will have to go through the new version code each time and check for new fields to update the boilerplate.

1notime0:
I think, I agree with @LeaveMyYard , the simple thing doesn't mean good thing.

LeaveMyYard (changes requested):
> So we will have to go through the new version code each time and check for new fields to update the boilerplate.

To be honest I think this is actually a problem. While boilerplate is not that critical, the need to rewrite it each time a new parameter is added to the codebase is really inconvenient. @aantn WDYT?

PR #220, opened 2024-03-02
Improve support for limited permissions

aantn:

PR #224, opened 2024-03-04
feat: cli option for --as-group

rohank07:
Add --as-group CLI flag if users are not using a group instead of a service account when using per-namespace features

part-of https://github.com/robusta-dev/krr/pull/220


aantn:
@rohank07 is it possible to add a file under tests with instructions for testing this? (To the degree that it is relevant, please copy the format in the other YAML file there for `--as`.)

PR #227, opened 2024-03-08
CSV Exporter

whaakman:
This exporter is based on the table exporter and can be called using the _-f csv_export_ parameter. The exporter stores a file in the current directory.

image


LeaveMyYard (approved):
Thanks for your contribution!

PR #231, opened 2024-03-14
Fix fileoutput

aantn:


dgdevops:
I have tested the changes from custom branch and both `--fileoutput` & `--slackoutput` flags generate the correct output when using table formatter.

LeaveMyYard (approved):

PR #235, opened 2024-03-17
use alias when available for strategy setting

maMykola:


maMykola:
it's possible to use pd.Field.alias if the change only needed for `allow_hpa` option

robusta_krr/main.py
```diff
@@ -282,7 +282,7 @@ def run_strategy(
                         kind=inspect.Parameter.KEYWORD_ONLY,
                         default=typer.Option(
                             field_meta.default,
-                            f"--{field_name}",
+                            f"--{field_meta.field_info.extra.get('param_decls') or field_name}",
```
LeaveMyYard, line 285: Is `param_decls` some documented thing in pydantic? Maybe better to use `alias`?

robusta_krr/strategies/simple.py
```diff
@@ -30,7 +29,11 @@ class SimpleStrategySettings(StrategySettings):
     points_required: int = pd.Field(
         100, ge=1, description="The number of data points required to make a recommendation for a resource."
     )
-    allow_hpa: bool = pd.Field(False, description="Whether to calculate recommendations even when there is an HPA scaler defined on that resource.")
+    allow_hpa: bool = pd.Field(
+        False,
+        param_decls="allow-hpa",
```
LeaveMyYard, line 34: It feels like it is a bit incorrect to have all parameters with `_` and one with `-`
Maybe a better thing to do is to have both? (for backwards-compatilibity) and then we can do some 
```.replace("_", "-")``` 
where we use those parameters, tho it will be a bit harder

LeaveMyYard (changes requested):
Thanks for your contribution! Left some messages to discuss
Can you please change the target branch to main, as allow-hpa is merged now

robusta_krr/main.py
maMykola: it won't work, I checked

robusta_krr/strategies/simple.py
maMykola: I did it before, but maybe it's better to make this change for major release?

maMykola: as it would be a BC

maMykola: ok, let me check if it's possible to have both options at the same time

maMykola: done

LeaveMyYard (approved):
Good job!

PR #239, opened 2024-03-21
Live introduction message

LeaveMyYard:
- Try to load the KRR introduction message from ~~git repository~~ Relay API (https://github.com/robusta-dev/relay/pull/76). Fallback to the local message file in case of something failed.
Additionally, move the ASCII art to that message (As I felt like it belongs there)
- Write that a newer KR version is available in case it is

@aantn Please feel free to update intro message

NOTE: Currently it is hard to show the message after the output, as we are parsing the last part of the message in a playbook (also mirrord has it in the beginning, so...)

### DIFF
--- a/robusta_krr/core/runner.py
+++ b/robusta_krr/core/runner.py
@@ -19,7 +19,7 @@ from robusta_krr.core.models.objects import K8sObjectData
 from robusta_krr.core.models.result import ResourceAllocations, ResourceScan, ResourceType, Result, StrategyData
 from robusta_krr.utils.intro import load_intro_message
 from robusta_krr.utils.progress_bar import ProgressBar
-from robusta_krr.utils.version import get_version
+from robusta_krr.utils.version import get_version, load_latest_version
 
 logger = logging.getLogger("krr")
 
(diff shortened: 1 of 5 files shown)


robusta_krr/core/runner.py
```diff
@@ -65,12 +65,35 @@ def _get_prometheus_loader(self, cluster: Optional[str]) -> Optional[PrometheusM
...
+
+    def __check_newer_version_available(self, current_version: str, latest_version: str) -> bool:
+        try:
+            current_version_parsed = self.__parse_version_string(current_version)
+            latest_version_parsed = self.__parse_version_string(latest_version)
+
+            if current_version_parsed < latest_version_parsed:
+                return True
+        except Exception:
+            logger.exception("An error occurred while checking for a new version")
```
Avi-Robusta, line 81: is an exception thrown if latest version is None?
If we are running in a cluster the IP might be blocked and this could always be printed
Will the logs from this be an issue with how runner formats the krr results?

Avi-Robusta: I guess this isnt run since we run it with -q, but I just want to check

Avi-Robusta (approved):
LGTM,
Just wanted to clairify that this wont be run in cluster

robusta_krr/core/runner.py
LeaveMyYard: You are actually correct, better write that as a debug message maybe

LeaveMyYard:
Update: this one will be reworked to query the message from the relay, not github

robusta_krr/core/runner.py
LeaveMyYard: No, actually we have a try catch in the function that gets the version
If it fails, it returns None

So this exception log (also not critical) is just because we do not expect to get a string with some invalid format here.

Still, I will make that a debug with a traceback, as someone at somepoint might set the invalid release version

PR #251, opened 2024-03-29
Load kubeconfig first in case it is provided by CLI arg

shlomosfez:
Currently, krr tries to load local cluster by default, and only if fails it tries to load from kubeconfig.

In this small PR, we first check if the kubeconfig or context arguments are provided, and if so, try to use it first

Thanks


aantn:
Thanks for the PR. As a slightly alternative implementation, would something like this work?

```
    def load_kubeconfig(self) -> None:
        try:
            config.load_kube_config(config_file=self.kubeconfig, context=self.context)
            self.inside_cluster = False
        except ConfigException:
            config.load_incluster_config()
            self.inside_cluster = True
```

It simplies the flow and should have the same behaviour in every case we care about. But note that I haven't tested it yet.

shlomosfez:
Love it! Changes applied

LeaveMyYard (approved):

PR #252, opened 2024-03-29
Add selectors to Prometheus discovery

bpfoster:
Adds labels to the promethues discovery selectors that follow the convention used by the prometheus-community chart.

https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/_helpers.tpl#L20
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/_helpers.tpl#L43

These are also the official Kuberneres [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).


aantn (approved):
LGTM. Thanks!

PR #253, opened 2024-03-29
Change selectors order - move the new selectors to be last, so existi…

arikalon1:
…ng behavior won't change for existing users, in case they have more than one match


arikalon1:
@bpfoster changed the selectors order, to prevent behavior change for existing users

aantn (approved):

PR #255, opened 2024-04-02
Make non-critical errors to be warnings

### LINKED ISSUE #213: `--fileoutput` writes to stdout instead of file
**Describe the bug**
It seems `--fileoutput` no longer works in 1.7.0 after https://github.com/robusta-dev/krr/pull/162

Instead of writing to file, it writes to stdout twice. The reason is most probably recursive call to [shadowed `print`](https://github.com/robusta-dev/krr/blob/72789f84e4ccad81f1549479bc9225e223671157/robusta_krr/utils/print.py#L8)

**To Reproduce**
1. Run KRR with `simple -f yaml --fileoutput <file>`
2. Recommendations are printed twice in stdout and not printed in file

**Expected behavior**
Recommendations are printed once in stdout and once in file

**Are you interested in contributing a fix for this?**
Yes

LeaveMyYard:
Closes #213 (error from discussion)

### DIFF
--- a/robusta_krr/core/runner.py
+++ b/robusta_krr/core/runner.py
@@ -208,7 +208,9 @@ class Runner:
         try:
             history_range = await prometheus_loader.get_history_range(timedelta(hours=5))
         except ValueError:
-            logger.exception(f"Was not able to get history range for cluster {cluster}")
+            logger.warning(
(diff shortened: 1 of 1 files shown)


arikalon1 (approved):
LGTM

PR #256, opened 2024-04-03
Refactor k8s workloads streaming

LeaveMyYard:
Remove streaming of workloads to simplify the code


aantn (approved):
LGTM. Before merging can you:

1. Look into the pytest failure
2. Verify on a test cluster that output on this branch is identical to output on main before this change?

Thanks!

LeaveMyYard:
Just checked: the only thing that slightly changes with this update is the ordering of workloads, which is OK.

(Previously could have been Deployment, Rollout, Statefullset, Daemonset, now it is Deployment, Rollout, Daemonset, Statefullset)

PR #258, opened 2024-04-04
Refactoring concept: generate typer commands from config

LeaveMyYard:
Related to #219

This is a proposal to move typer command generation to a separate file, restructure components in functions so it is easier to understand what is going on, and to use Pydantic's field extra params to generate typer commands from Config.

Pros: 
- Now to add a new flag it is only required to add it to config
- Was able to refactor the way a strategy is created, so now it is the same mechanism with strategy settings, so it is easier to understand
- Code is easier to read for newcomers (it is easier to understand how commands are generated)

Cons:
- We still use inspect in runtime


LeaveMyYard:
@aantn WDYT?

PR #268, opened 2024-04-26
Fix coralogix_token param

LeaveMyYard:


LeaveMyYard:
Closing, as we are afraid to break backwards compatibility for those who had everything working previously

PR #271, opened 2024-05-03
Add config to result

LeaveMyYard:
https://github.com/robusta-dev/robusta/pull/1397

### DIFF
--- a/robusta_krr/core/models/config.py
+++ b/robusta_krr/core/models/config.py
@@ -35,18 +35,18 @@ class Config(pd.BaseSettings):
 
     # Prometheus Settings
     prometheus_url: Optional[str] = pd.Field(None)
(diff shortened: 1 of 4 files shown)


robusta_krr/core/models/config.py
```diff
@@ -35,18 +35,18 @@ class Config(pd.BaseSettings):
 
     # Prometheus Settings
     prometheus_url: Optional[str] = pd.Field(None)
-    prometheus_auth_header: Optional[str] = pd.Field(None)
-    prometheus_other_headers: dict[str, str] = pd.Field(default_factory=dict)
+    prometheus_auth_header: Optional[pd.SecretStr] = pd.Field(None)
```
Avi-Robusta, line 38: I think this will break running it locally, if someone puts a normal string there get_secret_value might throw

Avi-Robusta (approved):

PR #277, opened 2024-05-09
Don't show HPA warning if user chooses to run w/ HPA

aantn:
I tested this by comparing output of the following two commands on this branch vs master and verifying that there were no unexpected changes to the output format other than what I wanted:

```
python3.9 krr.py simple --allow_hpa
python3.9 krr.py simple -f yaml --allow_hpa
```


arikalon1 (approved):
LGTM

PR #284, opened 2024-05-29
Security update

Avi-Robusta:
removed cves and updated base image


Dockerfile
```diff
@@ -1,16 +1,18 @@
 # Use the official Python 3.9 slim image as the base image
-FROM python:3.9-slim as builder
+FROM cgr.dev/chainguard/python:latest-dev as builder
```
Avi-Robusta, line 2: change base image and only copy whats needed to 
cgr.dev/chainguard/python:latest

arikalon1 (approved):
nice work

PR #291, opened 2024-06-15
Fix version command on dev builds

aantn:
# Motivation
Prior to this PR, `krr version` returned an incorrect hardcoded version number (1.8.2-dev) when run from source.

# Background
At present, when `krr version` is run from a proper release (e.g. installed with brew or downloaded from github) then it returns a correct version number that was patched by the github action for releases. When run from source, krr returns an incorrect number. 

We have a GitHub action which is supposed to patch the version upon release in the source code, but it doesn't work and hasn't for some time if ever. (E.g. [see a recent failure here](https://github.com/robusta-dev/krr/actions/runs/9463628920/job/26069236529).) I don't think that fixing that GitHub action is the correct approach either because it means we will have commits on main that never went through a PR. (E.g. commit generated by the auto-version-bump.)

Therefore I've removed that GitHub action and implemented a proper version command that takes dev releases into account.  The implementation is based on similar code I wrote for HolmesGPT.

# Testing
I tested the version command on a dev build and verified that it returns correct output.


arikalon1 (approved):
LGTM

PR #295, opened 2024-06-20
MAIN-1753 - patched pod conditions

Avi-Robusta:
This can be tested with

https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/controllers/job-pod-failure-policy-failjob.yaml


arikalon1 (approved):
nice work!

PR #296, opened 2024-06-20
Better --help text in krr

aantn:
# Motivation
Users often run `krr --help` to learn about KRR cli flags, but that command doesn't output most flags! Most of the flags are only shown for `krr simple --help`. This PR tells users where to look for the real cli flags.

# Alternatives
Ideally `krr --help` would show all cli flags, but I don't know how to do this with the library we use (typer) and I suspect it isn't possible. So while I would prefer doing that, this was the only solution I found.


pavangudiwada (approved):
LGTM

PR #312, opened 2024-07-08
fix: avoid exceed search.maxQueryLen

tlipoca9:


arikalon1 (approved):
nice work

thanks for the fix @tlipoca9

PR #330, opened 2024-08-29
fix: allow overriding mimir headers

davidspek:
When connecting to Mimir the `X-Scope-OrgID: anonymous` tenant header is automatically set. However, when using the `--prometheus-headers` flag to set a custom value for this it isn't propagated. This change ensures that headers set through the cli flags overwrite any `additional_headers` specified in the code.


chris-parker-za:
When can we expect this fix to be merged? Currently this is blocking mimir instances that don't use anonymous from working at all.

aantn (approved):
LGTM. Thank you!

aantn:
@chris-parker-za thank you for bumping this. Its merged and will be included in the next release.

chris-parker-za:
Happy to help! Glad it's approved finally.

PR #340, opened 2024-09-23
[MAIN-2169] upgrading the prometrix version

Avi-Robusta:
### Tests ran

- Ran krr before the upgrade and after the upgrade to compare the output


arikalon1 (approved):
LGTM

PR #352, opened 2024-10-26
Update README.md with minor fix for command to get bearer access toke…

dmealo:
…n for Azure Managed Prometheus


dmealo:
Realized the example was likely meant for bash rather than PowerShell (where it would not work to define a variable). Feel free to close or I can create a PowerShell example in addition. 🤷

aantn:
Yep, that's right! Closing for now. Thank you!

PR #353, opened 2024-10-27
add html formatter

saravanan30erd:
**Problem Statement**

Default option `table` provides pretty output in console using styles (colors, etc..) but when we save it into a file using `--fileoutput` styles will not be passed. If we share the output file to others via s3 or something, styles will be missed and looks plain. It would be nice to keep the styles when rendering from the file.

**Solution**

Default option `table` using `Rich` library and same library has function `export_html` to generate html from table output. It will keep the same styles used in default table formatter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
     <a href="https://github.com/robusta-dev/krr/issues">Request Feature</a>
     ·
     <a href="#support">Support</a>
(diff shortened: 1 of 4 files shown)


saravanan30erd:
@aantn Could you please take a look at this PR?
Thanks

aantn (approved):
Looks great. Very nice and clean PR.

PR #374, opened 2024-12-10
new simple cpu limit strategy 

RoiGlinik:
New KRR strategy.

the default simple strategy suggests to set cpu limits to none always.

Some users use multi-tenants clusters and would like a strategy where cpu limit can be set to specific cpu percentile.

New simple_limit strategy will support this use case. supports cpu request and limit percentiles and use them for the cpu recommendations.


nherment (approved):
This looks fine to me although I'm not very familiar with this part of the codebase.
Should there be some tests in addition? Although this change look simple, there are a few pieces of logic than can be tested.

PR #377, opened 2024-12-16
fix Argo Rollout CR wrong annotation type

Miracle05:
For Argo Rollouts managed objects we are getting annotations with type  `ObjectLikeDict` and the further object creation `obj = K8sObjectData` is falling with error 

`validation error for K8sObjectData annotations value is not a valid dict (type=type_error.dict) `

So we need to call a __dict__ method for explicit type conversion

### DIFF
--- a/robusta_krr/core/integrations/kubernetes/__init__.py
+++ b/robusta_krr/core/integrations/kubernetes/__init__.py
@@ -207,7 +207,10 @@ class ClusterLoader:
             labels = item.metadata.labels
 
         if item.metadata.annotations:
-            annotations = item.metadata.annotations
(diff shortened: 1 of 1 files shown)


kimsehwan96:
Not only `annotations` but also `labels` need same logic like below

```python3
        if item.metadata.labels:
            if type(item.metadata.labels) is ObjectLikeDict:
                labels = item.metadata.labels.__dict__
            else:
                labels = item.metadata.labels
```

RoiGlinik:
Thanks @Miracle05  and @kimsehwan96 . Will test it shortly and approve. Ill add the labels part myself .

RoiGlinik (approved):

PR #382, opened 2024-12-20
Add logics to change type when argocd rollout managed object' labels and annotations are passed

kimsehwan96:
Add https://github.com/robusta-dev/krr/pull/377 logic in labels.

Refactor these logics for more pythonic, and add `NOTE` comment.

### DIFF
--- a/robusta_krr/core/integrations/kubernetes/__init__.py
+++ b/robusta_krr/core/integrations/kubernetes/__init__.py
@@ -201,16 +201,15 @@ class ClusterLoader:
         namespace = item.metadata.namespace
         kind = kind or item.__class__.__name__[2:]
 
-        labels = {}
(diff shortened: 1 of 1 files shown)


aantn:
@kimsehwan96 is this still relevant?

kimsehwan96:
> @kimsehwan96 is this still relevant?

It was related with #377. There is missing logic for labels to solve that problem. So I added it and I did not follow up any changes after this pr submitted. If this issue is resolved than this PR can be closed without merge.

aantn:
Thanks. It's [been fixed on main](https://github.com/robusta-dev/krr/blob/main/robusta_krr/core/integrations/kubernetes/__init__.py#L216) already. Yeah, this fix some subtle improvments too.

PR #390, opened 2025-01-05
Fix csv/json output to file

### LINKED ISSUE #380: v1.18.0 produces invalid YAML on long lines
**Describe the bug**

**Take the version number with a grain of salt, we only noticed this due to the new feature of reporting labels and annotations**

When producing a yaml report with KRR that contains very long lines (seems to be pretty exactly 81 columns), krr adds a line break, which often breaks the yaml format.

Example:
`rewe-digital.com/scm-image-commit: <commit-hash>` produces a line that's longer than 80 lines. This is being output to the yaml files as
```
rewe-digital.com/scm-image-commit: 
 <commit-hash>
```
Which is invalid yaml, so it breaks programs trying to parse the file

**To Reproduce**
1. create some kubernetes resource where label + value is longer than 80 columns
2. analyze this resource with krr and configure yaml output
3. check the produced yaml file

**Workaround**
For now simply downgrading to 1.17.0 fixes the issue since the labels and annotations are not being written to the output report

aantn:
Fixes #380

### DIFF
--- a/robusta_krr/core/runner.py
+++ b/robusta_krr/core/runner.py
@@ -121,7 +121,7 @@ class Runner:
 
             with open(file_name, "w") as target_file:
                 # don't use rich when writing a csv or html to avoid line wrapping etc
-                if settings.format == "csv" or settings.format == "html":
+                if settings.format == "csv" or settings.format == "html" or settings.format == "json" or settings.format == "yaml":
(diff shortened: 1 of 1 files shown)


arikalon1 (approved):
LGTM

PR #388, opened 2025-01-01
feat: add support for StrimziPodSet

sappusaketh:
fixes #372

PR #391, opened 2025-01-08
Fix error on labels parsing  for ArgoRollout objects

YGwen:
An Error occurs on Rollout when parsing the labels because the time of labels apparently sometimes is robusta_krr.utils.object_like_dict.ObjectLikeDict

`
File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__ 
pydantic.error_wrappers.ValidationError: 1 validation error for K8sObjectData
labels
value is not a valid dict (type=type_error.dict) 
`

This PR re-uses the fix that was applied on annotations and fixes the issue.


aantn:
Thanks. We've already fixed in https://github.com/robusta-dev/krr/pull/388 so closing this

PR #407, opened 2025-02-24
Fix bug with AWS Credentials

aantn:
When the user overrides AWS credentials via CLI, don't try to fetch credentials via boto3 (this can fail and isn't necessary anyway)


arikalon1 (approved):
LGTM

PR #408, opened 2025-03-04
make slack api safer and update to file_upload_v2

Avi-Robusta:
origonal file upload api is deprecated
due to limitations of file_upload_v2 the message` Kubernetes Resource Report for splunk` needs to be sent seperatly since it is used to get the internal channel ID name which file_upload_v2 requires

Looks identical - first is orig, second is new
Screen Shot 2025-03-04 at 16 33 54


nherment (approved):

Avi-Robusta:
merged in other branch

Issue #415, opened 2025-03-20
Scan pod alone


mysiki:
**Is your feature request related to a problem? Please describe.**
My problem is link to the usage of Vcluster. With Vcluster, I have virtual cluster (with proper etcd) inside my K8S cluster. The workload (pod) are run / sync on host cluster but not the other component like Deployement, STS, etc ...
So KRR is not able to scan those pod because KRR only focus on Deployement, sts etc. 

Prom exporter is on host cluster, so only know objects present on host cluster (not inside the vcluster). 

Bonus, to avoid conflict, Vcluster rename the pod during the sync. So Prom have metric for pod like "myPod-HashId" when inside vcluster the pod name is "myPod". That mean, event running KRR inside vcluster is not possible because the POD name (inside vcluster) is not the same as the metric name that come from Host cluster. 

**Describe the solution you'd like**
Ability to scan POD object directly.

**Describe alternatives you've considered**
Run inside vcluster, but as name are different, this not work. 

**Are you interested in contributing a PR for this?**
Yes. (but my python knowledge are really low :D) If yes, we will provide guidance what parts of the code to modify and help you.

**Additional context**
Not sure if this PR cover my needed : https://github.com/robusta-dev/krr/pull/266

Example from Host view (where prom metric are) : dp-app-demo-5fbc4d8f54-2cm9c-x-demo-develop-client1--073b2da98c
Same inside vcluster : dp-app-demo-5fbc4d8f54-2cm9c

macrokernel:
This would also be beneficial for such operators as Strimzi Kafka or Cloudnative-PG, which manage pods directly, without a deployment/ds/sts.

PR #421, opened 2025-03-28
Feature/vcluster

mysiki:
Add support of VCluster (https://www.vcluster.com/) (resolve my issue : https://github.com/robusta-dev/krr/issues/415)

VCluster (VC) work on top of physical cluster with workload sync between virtual cluster and physical cluster. During the sync, VC rename the pod to avoid naming collision. VC only sync the pod, not the other stuff like RS, Deployment, ... The sync is done on the physical namespace where vcluster is present. 

Summarize design : 

| Physical cluster Namespace | Physical cluster pod name | Virtual cluster Namespace | Virtual cluster pod name |
|--|--|--|--|
| vcluster-foo | myappfront-1234-x-myappns-x-vclusterfooname | myappns | myappfront-1234 |
| vcluster-foo | myappback-123-x-myappns-x-vclusterfooname | myappns | myappback-123 |
| vcluster-foo | otherapp-1234-x-otherappns-x-vclusterfooname | otherappns | otherapp-1234 |
| vcluster-foo | otherapp-4567-x-otherappns-x-vclusterfooname | otherappns | otherapp-4567 |
| vcluster-foo | otherappwithreallylongnammmmme-1234567891234567891234-x-othera-f5fef5e6fe | otherappns | otherappwithreallylongnammmmme-1234567891234567891234 |

The renaming is done following algo present in added function get_vcluster_pod_real_name.

To reduce pod number and cluster load, usually metrics are read read from physical cluster (avoid to have all exporter pod inside VC). 

Because of VC only sync pod, KRR need to run connected to the VC (to detect DP, STS,..). But the pod name and pod namespace in prometheus are the one present in physical cluster.

This PR add 2 new arguments that are mandatory to calculate the pod name and pod namespace for prom query and function to use it.
I also add debug logger on prom query (can be remove if you think is useless :) ) 

Readme is update

Test done with success on my VC (with VC args) and without (like before this change).


mysiki:
some news ?

PR #438, opened 2025-06-09
ROB-1415 update prometrix

RoiGlinik:


Avi-Robusta (approved):
tested in cluster, no issues

PR #442, opened 2025-06-26
[ROB-1588] Push results to runner on success or failure

Avi-Robusta:
tested in cluster and cli
in cluster verified errors in multiple places in code are sent to scan failure in the db


robusta_krr/core/runner.py
```diff
@@ -348,9 +353,64 @@ async def run(self) -> int:
...
             return 0  # Exit with success
+
+    def _send_result(self, url: str, start_time: datetime, scan_id: str, result: Result):
+        result_dict = json.loads(result.json(indent=2))
+        _send_scan_payload(url, scan_id, start_time, result_dict, is_error=False)
+
+def publish_error(url: str, scan_id: str, start_time: str, error: str):
+    _send_scan_payload(url, scan_id, start_time, error, is_error=True)
+
+def publish_error(error: str):
```
moshemorad, line 372: Are you sure that this will work? e.g. i don't think that python support method overloading that way.

robusta_krr/core/runner.py
```diff
@@ -348,9 +353,64 @@ async def run(self) -> int:
...
+            "scan_id": scan_id,
+            "start_time": start_time,
+        }
+    }
+
+    try:
+        logger_msg = "Sending error scan" if is_error else "Sending scan"
+        logger.warning(logger_msg)
+
+        response = requests.post(
```
moshemorad, line 404: Do we want to have some retry machnisam here?

moshemorad (changes requested):

moshemorad (changes requested):

PR #443, opened 2025-06-28
krr-enforcer: Support envFrom in helm values

k0da:
For the cases like:
```ENV var replacement ROBUSTA_UI_TOKEN  does not exist for param: {{ env.ROBUSTA_UI_TOKEN }}```
Robusta runner sources Token via secret and refers it through envFrom. In order to inforcer
re-use secret, this commit adds envFrom support.

For example:
```
envFrom:
  - secretRef: name: robusta-runner-secret optional: true
```

### DIFF
--- a/helm/krr-enforcer/templates/enforcer.yaml
+++ b/helm/krr-enforcer/templates/enforcer.yaml
@@ -111,6 +111,10 @@ spec:
               memory: {{ .Values.resources.limits.memory }}
               {{- end }}
             {{- end }}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by k0da** [comment]
@arikalon1 suggested alternative way using existing codebase:
```yaml
additionalEnvVars:
  - name: ROBUSTA_UI_TOKEN
    valueFrom:
      secretKeyRef:
        name: "my-secret"
        key: uiToken
```

Feel free to close.

PR #446, opened 2025-07-12
feat: add CPU limit multiplier to simple limit strategy

patsevanton:
This PR introduces a new configuration parameter `cpu_limit_multiplier` to the SimpleLimitStrategy, providing more flexible control over CPU limit recommendations.

**Changes:**
- Added `cpu_limit_multiplier` field to `SimpleLimitStrategySettings` with default value of 1.0
- Updated CPU limit calculation to multiply the percentile-based value by the multiplier
- Modified strategy description to reflect the new multiplier in the output format
- Updated example command in description to include the new parameter

**Benefits:**
- Allows users to set CPU limits as a percentage of the calculated percentile (e.g., 200% of 96th percentile)
- Provides more granular control over resource allocation strategies
- Maintains backward compatibility with default multiplier of 1.0
- Useful for workloads that need headroom above statistical percentiles

**Example Usage:**
```bash
krr simple_limit --cpu_limit_multiplier=2.0  # Sets limit to 200% of percentile
krr simple_limit --cpu_limit_multiplier=1.5  # Sets limit to 150% of percentile
```


aantn:
Isn't that what you're trying to achieve though?

PR #451, opened 2025-07-23
Feature/azureblob-output

mlhv:
Added option to send report to Azure Blob Storage, as well as optionally notify a Teams Channel provided a webhook URL. If using the teams webhook alert, then an `azure-subscription-id` and `azure-resource-group` is required.

Users can use the following new flags when running the program:

- `--azurebloboutput "Blob Service SAS URL Here" `
- `--teams-webhook "Teams Webhook URL Here" `
- `--azure-subscription-id "Azure Subscription ID Here" `
- `--azure-resource-group "Azure Resource Group Name Here" `

Documentation is added to README.


aantn (dismissed):
Thank you!

arikalon1 (approved):
nice work

Thanks for this great PR @mlhv

PR #457, opened 2025-07-31
Ability to get the related pod label name via an environment variable.

StFS:
This PR adds the ability to specify the KRR_RELATED_POD_LABEL environment variable to provide the name of the label to use to discover a K8S object related pods. The default behavior, if this env variable is not set, is to use `pod` as before.

This is helpful for example when the `honor_labels` setting in Prometheus is set to `false` (the default value) and there is a conflict between labels that are being scraped and labels that Prometheus attaches.
See documentation here: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

This can result in the metrics using the label `exported_pod` instead of `pod`.


arikalon1 (approved):
nice and clean

Thanks for the PR @StFS

PR #466, opened 2025-08-26
Add an env var that allows excluding containers by name

arikalon1:
Useful for containers that have memory spike that is too fast to be captured by krr (prometheus)


Avi-Robusta (approved):
LGTM

PR #525, opened 2026-05-20
feat: add Python 3.13 support

patsevanton:
### Description

This PR adds Python 3.13 compatibility to the project while maintaining backward compatibility with Python 3.10–3.12.

### Changes

**pyproject.toml**
- Expand Python version range from `>=3.10,<=3.12.9` to `>=3.10,<3.14`
- Relax numpy version constraint from `>=1.26.4,<1.27.0` to `>=1.26.4,<3` (numpy 1.26.x does not support Python 3.13)
- Relax pandas version constraint from `==2.2.2` to `>=2.2.3,<4` (pandas 2.2.2 fails to compile on Python 3.13)
- Upgrade typer from `^0.7.0` to `>=0.9.0,<1` (typer 0.7.0 is incompatible with click 8.4+ required by Python 3.13)

**robusta_krr/main.py**
- Fix `show_severity` option declaration: change `" /--exclude-severity"` to `"--show-severity/--exclude-severity"` — the old syntax caused `TypeError: Secondary flag is not valid for non-boolean flag` with newer click versions
- Add `None` guard for `namespaces` and `resources` before checking `"*" in value` to prevent `TypeError` when these options are not provided

**tests/test_krr.py, tests/test_runner.py**
- Remove `mix_stderr=False` argument from `CliRunner()` — this parameter was removed in newer typer versions


patsevanton:
@aantn Please review the changes and provide feedback.

Issue #411, opened 2025-03-12
Customize metrics for memory and cpu


jonpalle:
**Is your feature request related to a problem? Please describe.**
I have access to and would like to get recommendations about resources running in various namespaces, but I do not have access to Kubernetes node metrics. I have used node-exporter to expose CPU and memory metrics in Grafana via open telemetry collector and would like to use those metrics for KRR to give me recommendations.

**Describe the solution you'd like**
An option for `krr simple` enabling me to set the metrics used for CPU and memory. E.g. to use `node_memory_MemAvailable_bytes` instead of `container_memory_working_set_bytes`. Currently, I have a Grafana panel that uses `100 * (1 - (node_memory_MemAvailable_bytes{app="$APP"} / node_memory_MemTotal_bytes{app="$APP"}))` to show me the memory usage in percent.

**Describe alternatives you've considered**
The only alternative that I have considered is to ask the administrators of my Kubernetes cluster for permission to access the node metrics, but I doubt that will happen.

**Are you interested in contributing a PR for this?**
Yes.

**Additional context**
I will be happy to add any more context needed for this.

Issue #354, opened 2024-10-31
Required metrics in README is missing a few values


voidlily:
**Describe the bug**
In my cluster I'm having issues with the high cardinality of cAdvisor metrics and trying to make an allowlist of only the metrics krr needs. The readme mentioned these metrics

```
container_cpu_usage_seconds_total
container_memory_working_set_bytes
```

I noticed however that there were others missing like `machine_memory_bytes` and `machine_cpu_cores` that the cluster summary asks for and are otherwise missing.

For kube-state-metrics there's a few others I noticed but might have missed, because I haven't yet tried limiting kube-state-metrics because its cardinality was not as high as cAdvisor was.

```
kube_pod_container_resource_limits
kube_job_owner
kube_pod_container_status_last_terminated_reason
```

This might not even be all of them, I didn't search very deeply in the code yet.

**Expected behavior**
A list of all metrics that krr expects so I can make a thorough allowlist. Possibly tailored to which component of kube-prometheus-stack each comes from (cAdvisor, kubelet, kube-state-metrics, etc), though not strictly necessary.

aantn:
Thanks, you're right! If you're open to contributing, we'd love to accept a PR improving the docs on this.

PR #266, opened 2024-04-22
Prometheus workload loader

LeaveMyYard:

PR #546, opened 2026-07-31
Support OpenTelemetry metrics alongside kube-state-metrics

ElfoLiNk:
## Problem

KRR hardcodes kube-state-metrics / cAdvisor / node-exporter metric names. Clusters whose Kubernetes **state** metrics come from an OpenTelemetry Collector (`k8s_cluster` + `hostmetrics` receivers) expose the same information under different names, so KRR finds no pods, no node capacity and no OOMKills there. Disabling kube-state-metrics is common in OTel-based observability stacks.

## What this adds

**1. A metric dialect for Kubernetes state metrics** — `--prometheus-metrics-dialect {auto,kube-state-metrics,otel}`

| KRR needs | `kube-state-metrics` | `otel` | OTel source |
|---|---|---|---|
| node memory | `machine_memory_bytes` (by `instance`) | `system_memory_limit_bytes` (by `host_name`) | hostmetrics, `system.memory.limit` |
| node CPU | `machine_cpu_cores` (by `instance`) | `system_cpu_logical_count` (by `host_name`) | hostmetrics, `system.cpu.logical.count` |
| memory requests | `kube_pod_container_resource_requests{resource="memory"}` | `k8s_container_memory_request_bytes` | k8s_cluster |
| CPU requests | `kube_pod_container_resource_requests{resource="cpu"}` | `k8s_container_cpu_request` | k8s_cluster |
| memory limits | `kube_pod_container_resource_limits{resource="memory"}` | `k8s_container_memory_limit_bytes` | k8s_cluster |
| pod phase | `kube_pod_status_phase{phase="Running"} == 1` | `k8s_pod_phase == 2` (phase enum) | k8s_cluster |
| OOMKills | `kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}` | `k8s_container_status_reason{k8s_container_status_reason="OOMKilled"}` | k8s_cluster, `k8s.container.status.reason` (off by default) |

Container **usage** metrics (`container_cpu_usage_seconds_total`, `container_memory_working_set_bytes`) are deliberately *not* part of the dialect — they keep their cAdvisor names in every setup, so an OTel-only deployment still scrapes cAdvisor (e.g. with the Collector's `prometheus` receiver). This keeps the change scoped to state metrics.

**2. Pod owner resolution as a separate axis** — `--prometheus-owner-resolution {auto,kube-state-metrics,recording-rule}`

The `k8s_cluster` receiver has no equivalent of the `kube_*_owner` metrics, so this cannot be expressed as a name swap.

- `kube-state-metrics` — the existing owner chain, unchanged, including the per-kind handling of Rollout / DeploymentConfig / CronJob / GroupedJob and `KRR_OWNER_BATCH_SIZE` batching.
- `recording-rule` — one query against a pod owner recording rule (`namespace_workload_pod:kube_pod_owner:relabel` by default, override with `--prometheus-workload-recording-rule`). Also saves a query per workload on large clusters. It only covers Deployment, DaemonSet, StatefulSet, Job and ReplicaSet, so other kinds fall back to the owner chain rather than silently returning no pods.

## Backwards compatibility

Both `auto` modes probe kube-state-metrics **first** and cache the result per service, so an existing installation keeps sending exactly the queries it sent before, at the cost of one `count()` probe per cluster. If nothing is detected, KRR logs a warning and falls back to kube-state-metrics. The only intentional change to a kube-state-metrics query is the quote style in the kube-system request queries (`'kube-system'` → `"kube-system"`).

## Tests

`tests/test_metric_dialects.py` (18 tests) pins the generated PromQL for both dialects, using the pre-existing kube-state-metrics queries as golden values, and covers detection precedence, caching, explicit selection, and the owner-resolution fallbacks. Full suite: 73 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Read more about [how KRR works](#how-krr-works)
 
 ### Requirements
 
-KRR requires Prometheus 2.26+, [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) & [cAdvisor](https://github.com/google/cadvisor).
+KRR requires Prometheus 2.26+, [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) & [cAdvisor](https://github.com/google/cadvisor). cAdvisor is always required; kube-state-metrics can be replaced by an OpenTelemetry Collector, see [Using OpenTelemetry instead of kube-state-metrics](#requirements).
 
 <details>
   <summary>Which metrics does KRR need?</summary>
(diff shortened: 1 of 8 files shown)


ElfoLiNk:
Cross-references to existing issues/PRs, for whoever reviews this:

**Overlaps with #420** — that PR changes the same `get_cluster_summary` lines, replacing the deprecated `machine_memory_bytes` / `machine_cpu_cores` with `kube_node_status_capacity{resource="..."}` (they are missing on Azure Managed Prometheus). It also makes the same `'kube-system'` → `"kube-system"` quote change this PR makes, so the two will conflict textually.

I am happy to rebase on top of #420 in whichever order you prefer. Worth noting: #420 also drops the `max by (instance)` wrapper, which the dialect cannot express with just a metric name and a group-by label, so folding it in means either adding a resource-selector field to `MetricDialect` or keeping the node query shape per dialect. Say which you'd rather have and I'll adjust.

**Partially answers #354** (complete list of required metrics) — the new README section tabulates every Kubernetes state metric KRR queries, in both dialects, with the OpenTelemetry receiver each comes from.

**Related but not closed by this PR:**

* #411 — requests configurable CPU/memory metric names for a node-exporter-behind-OTel-collector setup. This PR deliberately leaves container *usage* metrics on their cAdvisor names, so that request stays open.
* #59 / #266 — Prometheus-based workload discovery. The `kube-state-metrics` owner-resolution path here walks the same owner chain, so there may be code to share if #266 moves forward.

**If #512 or #323 (pydantic v2) lands first**, `MetricDialect`'s `class Config: allow_mutation = False` becomes `model_config = ConfigDict(frozen=True)`. One line.

ElfoLiNk:
Updated to fold #420 in, so the two PRs no longer collide over the same four lines.

The `kube-state-metrics` dialect now reads node capacity from `kube_node_status_capacity{resource="memory"|"cpu"}` grouped by `node`, instead of the deprecated `machine_memory_bytes` / `machine_cpu_cores` grouped by `instance`. This is a behaviour change on the default path, deliberately: those metrics are missing on Azure Managed Prometheus, which is what #420 reports.

Difference from #420: the `max by (node)` wrapper is kept rather than a bare `sum()`, so a node scraped by more than one job is still counted once.

Also expanded the README requirements list to name every metric KRR queries and the component that exposes it, which is what #354 asks for.

74 tests pass, including a new case asserting the cluster-summary matchers stay correct when `-l/--prometheus-cluster-label` is set.

ElfoLiNk:
Went through the CodeRabbit review. Summary of what changed and what I pushed back on.

**Accepted — the label mapping (`metric_dialects.py` finding).** The example in that comment pointed at `memory_resource_selector=""`, which is intentional (OTel encodes the resource in the metric name), but the underlying point is correct and important: the `otel` dialect assumed the Kubernetes resource attributes are exposed as `namespace` / `pod` / `container`. A Collector that exports them verbatim produces `k8s_namespace_name` / `k8s_pod_name` / `k8s_container_name`, which cannot be joined with the cAdvisor usage metrics — and KRR keys its results on `pod`, so it would have failed with a `KeyError` rather than a clear message. My test cluster only worked because its pipeline has a `transform` processor doing that mapping, which I had not noticed.

Now: the requirement is documented with the `transform` processor that implements it, dialect detection probes for the mapped label rather than only the metric name, and a pipeline that has `k8s_pod_phase` without `namespace` is reported explicitly instead of being accepted. Test: `test_otel_is_rejected_when_the_labels_are_not_mapped`.

**Accepted — probe over a lookback window.** Detection now uses `count(last_over_time(<metric>{...}[1h]))`, window overridable via `KRR_DIALECT_PROBE_WINDOW`. A silent scan-wide dialect switch after a scrape gap was the right thing to worry about. Test: `test_probes_use_a_lookback_window`.

**Accepted — all four README findings.** kube-state-metrics is now stated as replaceable (cAdvisor is not), the recording rule's required labels include `namespace` and the configured cluster label, the `auto` fallback chain is described as implemented, and the historical-pod note names the three metrics instead of saying "last three".

**Unrelated fix found while addressing the above:** `gather_data` read `metric_loader.service_name` in its no-data warning, but `metric_loader` is unbound when the constructor raises — that path was a `NameError`. Now uses `self.name()` / `LoaderClass.__name__`.

Also verified against a live OTel/Thanos stack (`k8s_cluster` + `hostmetrics` + cAdvisor via the `prometheus` receiver): dialect and owner resolution auto-detect correctly, 31 deployments scanned, no warnings. Cluster summary values check out (437 GB / 52 cores / kube-system 15.6 GB and 7.02 cores).

One data point for #420 from that stack: `system_cpu_count` — which I had used in an earlier revision — is the Micrometer JVM metric, not a node metric. It reports 128 across 63 JVM series versus 52 actual logical cores on 13 nodes. The dialect uses `system_cpu_logical_count`.

77 tests pass.

ElfoLiNk:
Both of the latest two findings accepted, pushed as `a63a9ae`.

**Warn when neither owner source has data** — correct, and it was asymmetric with `_detect_dialect`, which already warns in the same situation. An OpenTelemetry-only cluster without the recording rule would have produced empty pod lists for every workload with nothing naming the cause. Now warns and says what it means for the results (recommendations limited to running pods). Covered by `test_owner_resolution_defaults_to_owner_metrics`.

**`--prometheus-workload-recording-rule` help text** — correct, it disagreed with the README. Now lists `workload`, `workload_type`, `pod`, `namespace` plus the cluster label.

77 tests pass, black/flake8 clean under the repo configs.

